PR #5, opened 2021-12-23
port accession ID increment code to phylo tree

morsecodist:
Applied your code from the consensus genome here.


rzlim08:
I think this breaks one of the tests for some reason

morsecodist:
@rzlim08 fixed thanks!

rzlim08 (approved):

PR #20, opened 2022-02-01
longest reads

morsecodist:
Output longest reads in alignment viz for blast.


ovalenzuela19 (approved):
Thanks for doing this!!

PR #22, opened 2022-02-04
add WDLs for alignment

morsecodist:
For the download cache we can actually leverage our miniWDL batch job definition which already has this implemented. Here I am creating WDL versions of our alignment steps.


rzlim08:
Cool! Will this mean we can get rid of the `entrypoint.sh` code that lives in the monorepo?

morsecodist:
@rzlim08 yep and the docker files in there, this actually makes the whole thing cleaner I probably should have thought of this before

rzlim08:
> @rzlim08 yep and the docker files in there, this actually makes the whole thing cleaner I probably should have thought of this before

It looks good, only thing is the Diamond -> Minimap2 name change

rzlim08 (approved):

PR #25, opened 2022-02-11
alignment miniwdl

morsecodist:
Refactors alignment to unify some code and run them as miniwdl wdls.


short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -1,77 +1,106 @@
...
+
+import boto3
 from botocore.exceptions import ClientError
 
 log = logging.getLogger(__name__)
 
 MAX_CHUNKS_IN_FLIGHT = 10
+ALIGNMENT_WDL_VERSIONS: Dict[str, str] = {
+    "diamond": "v1.0.0",
+    "minimap2": "v1.0.0",
```
rzlim08, line 25: Eventually it might be good to have these in some sort of parameter store where we can increment as we release new versions

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -1,77 +1,106 @@
...
-        ).json()["accountId"]
-    return f"aegea-batch-jobs-{account_id}"
+_batch_client = boto3.client("batch")
+_s3_client = boto3.client("s3")
+
+try:
+    account_id = boto3.client("sts").get_caller_identity()["Account"]
+except ClientError:
+    account_id = requests.get(
+        "http://169.254.169.254/latest/dynamic/instance-identity/document"
```
rzlim08, line 36: Is it possible to use IMDSv2 for this?

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -1,77 +1,106 @@
...
+    batch_job_desc_bucket = boto3.resource("s3").Bucket(f"aegea-batch-jobs-{account_id}")
     key = f"job_descriptions/{job_id}"
     try:
         job_desc_object = batch_job_desc_bucket.Object(key)
         return json.loads(job_desc_object.get()["Body"].read())["status"]
     except ClientError as e:
         if e.response["Error"]["Code"] == "NoSuchKey":
             # Warn that the object is missing so any issue with the s3 mechanism can be identified
-            log.debug("missing_job_description_ojbect", extra={key: key})
+            log.debug(f"missing_job_description_ojbect key: {key}")
```
rzlim08, line 58: nit: spelling of "object"

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -84,46 +113,128 @@ def _run_batch_job(job_name, job_queue, job_definition, environment, chunk_id, a
...
+    chunk_id: int,
+    db_chunk: str,
+):
+    deployment_environment = os.environ["DEPLOYMENT_ENVIRONMENT"]
+    pattern = r"s3://.+/samples/([0-9]+)/([0-9]+)/"
+    m = re.match(pattern, input_dir)
+    if m:
+        project_id, sample_id = m.group(1), m.group(2)
+    else:
+        project_id, sample_id = "0", "0"
```
rzlim08, line 148: it actually might be better to fail here to avoid confusion

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -1,77 +1,106 @@
...
+
+    def _log_status(status: str):
+        level = logging.INFO if status != "FAILED" else logging.ERROR
+        log.log(
+            level,
+            "batch_job_status " + json.dumps({
+                "job_id": job_id,
+                "job_name": job_name,
+                "job_queue": job_queue,
+                "job_definition": job_definition,
```
rzlim08, line 98: It might be good to print the environment, since failures often come from incorrect environment inputs

rzlim08 (changes requested):
Looks good! Just a couple of nits. 

Do you think we should check the mm2/diamond wdl files into this repo? Also, not sure how you feel about running `black` on the file. I like to do that for readability, but that's totally optional.

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
morsecodist: While this way is clunky I actually think hard-coding it is the best we can do right now. It is important that the version of the overall pipeline should include the versioning of these pipelines. So for example v6.11.0 of the main pipeline would correspond to v1.0.0 of the aligners. To enforce this we would need to release the aligners then explicitly upgrade them as a change to the main pipeline.

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
morsecodist: Probably, let me see

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
morsecodist: That would preclude us from running ad-hoc workflows not from the application.

morsecodist:
@rzlim08

> Do you think we should check the mm2/diamond wdl files into this repo?

I already have https://github.com/chanzuckerberg/czid-workflows/blob/main/minimap2/minimap2.wdl https://github.com/chanzuckerberg/czid-workflows/blob/main/diamond/diamond.wdl

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
rzlim08: Ah good point

rzlim08:
> @rzlim08
> 
> > Do you think we should check the mm2/diamond wdl files into this repo?
> 
> I already have https://github.com/chanzuckerberg/czid-workflows/blob/main/minimap2/minimap2.wdl https://github.com/chanzuckerberg/czid-workflows/blob/main/diamond/diamond.wdl

Right, I forgot about that, thanks!

rzlim08 (approved):

PR #28, opened 2022-02-15
fix longest reads

morsecodist:
Fix longest reads bug and switch to the fasta format


ovalenzuela19 (approved):
Thank you so much for making this change! 🙏

PR #34, opened 2022-03-23
cleanup cache downloads

morsecodist:
Fix call caching by removing the downloaded cache dir from the chunks directory for merging. It is a little inefficient to download stuff just to delete but these files will be quite small so I think the complexity of a more complicated downloader that skips these files isn't justified.


morsecodist:
I am pretty sure the test failures are unrelated to this PR

rzlim08 (approved):

PR #35, opened 2022-03-31
Fix local benchmarks

rzlim08:
* Updating local benchmarks to align with minimap2/diamond outputs


rzlim08:
* Had to pin the jupyter scipy notebook version for compatibility reasons

mlin:
The `_LAST` symlink looks like an artifact of a miniwdl run & shouldn't be in git

rzlim08:
> The `_LAST` symlink looks like an artifact of a miniwdl run & shouldn't be in git

Good catch, thanks!

mlin (approved):

PR #38, opened 2022-04-14
wdl for index generation + phage list

morsecodist:
This PR refactors index generation into a WDL, implements index generation four alignment scalability based on @rzlim08 's PR, and generates a phage list. This should make index generation more testable and maintainable in the long run. I also ported over tests, though most of the old tests were just smoke tests checking for outputs which is done automatically by miniwdl here.


rzlim08 (changes requested):
I pulled this, ran the tests, and everything looks good, although I'm not so familiar with the expected structure of the taxon files. 

Just a couple of changes to the minimap2 script that changed from my PR to production code.

rzlim08 (approved):
Looks good!

PR #43, opened 2022-04-27
optimize index generation

morsecodist:
A few quick optimizations:

- Add `is_phage` to the `taxon_lineages` table. This is both easier to generate and easier for the webapp to use.
- Break down downloads into four separate jobs to both parallelize them and start downstream jobs ASAP


rzlim08 (approved):
Looks good

PR #46, opened 2022-05-04
host genome wdl

morsecodist:
Create a WDL workflow for host genome generation.


rzlim08 (approved):
This looks good, but where do the files end up? If we pass in the `s3_wdl_uri` variable, would we be able to upload directly to the czid-public-references dir?

PR #49, opened 2022-05-06
Increase retry count to 5 for spot, 3 for ec2

rzlim08:
* IMO we should still consider converting this call to a batch array, but this should be a reasonable mitigation
* Should lessen the "TooManyRequest" failures that we get

### DIFF
--- a/short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
+++ b/short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
@@ -28,7 +28,7 @@ ALIGNMENT_WDL_VERSIONS: Dict[str, str] = {
 }
 
 config = Config(
-    retries = {
+    retries={
         "max_attempts": 10,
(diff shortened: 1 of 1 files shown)


morsecodist (changes requested):
I don't think this retry is the issue. These are the number of retries in the case where the job fails. In he case of the too many requests exception the jobs are not failing, we are hitting is the API request limit. I think what you are looking for is configuring the batch client to retry the API request more times. See the [docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html) on that.

I think we would need something more like:

```python

from botocore.config import Config

config = Config(
   retries = {
      'max_attempts': 10,
      'mode': 'standard'
   }
)

_batch_client = boto3.client("batch", config=config)
```

There is already a default retry of 5. With the standard mode we are capped at 20 seconds, the legacy gives us unlimited time but recognizes fewer errors. If the built ins are too limiting we may want to write a simple custom exponential backoff.

PR #53, opened 2022-05-11
Upgrade boto

rzlim08:
* We couldn't use adaptive mode in boto because we were using an outdated version. This upgrades the version.


morsecodist (approved):

rzlim08:
Closing as I don't think this is necessary anymore

PR #54, opened 2022-05-12
ignore warnings for diamond

morsecodist:
Some sequences in NR are only composed of DNA characters (ACTG). This causes makedb to fail as a safeguard because it appears that we are trying to use it on DNA instead of protein sequences. I inspected these sequences and it seems like in all of NR there do happen to be some sequences of only these characters. They are all valid protein characters as well. With such a large and imperfect data set running into these sequences is inevitable so I think we should ignore warnings.

### DIFF
--- a/index-generation/index_generation.wdl
+++ b/index-generation/index_generation.wdl
@@ -223,7 +223,8 @@ task GenerateIndexDiamond {
     }
 
     command <<<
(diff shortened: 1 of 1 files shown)


rzlim08 (approved):
Looks Good!

PR #56, opened 2022-05-13
Split GenerateIndexAccessions into substeps for more performance

morsecodist:
It actually turns out that we made our alignment indexes so fast that Accessions is the bottleneck. I remember back when alignment index generation took three days the 12 + hours of accessions seemed hardly worth worrying about. This is a super low hanging fruit optimization i realized when I studied the step. There is no reason to block generating the `_loc.db` and `_info.db` files on accessions so we can parallelize them easily by splitting them out into their own steps.

Sadly, generating the accession DB itself is still quite slow but optimizing that presents much more of a challenge. I found a parallelism setting in the code and I doubled it from 8 to 16 (we have 60 cores to work with so that should be good). However, the parallelism is only for the actual generation, I suspect it spends a lot of it's time in writing the output files to the Berkley DB format, which is not parallelized. The print statements were hidden because they write to stdout and miniwdl only logs from stderr so I switched to use logging and added a log before the write step so we can look at the logs to get an idea of how long it spends on each element. I think with minimal work we could dramatically increase the performance of generating the accession index. There is no reason this should take 10+ hours to run.

### DIFF
--- a/index-generation/generate_accession2taxid.py
+++ b/index-generation/generate_accession2taxid.py
@@ -47,9 +47,8 @@ import argparse
 import dbm
 import gzip
 import shelve
+import logging
 from multiprocessing.pool import ThreadPool
-
-NUM_PARTITIONS = 8
 
 
 def output_dicts_to_db(mapping_files, wgs_accessions,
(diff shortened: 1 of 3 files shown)


rzlim08 (approved):
LGTM!

PR #63, opened 2022-05-19
Index turbo

morsecodist:
This is a sort of kitchen sink of improvements to index generation:
- Switch to using pigz for un-gzipping all of our downloads. Our python scripts were doing some uinzipping, this should be WAY slower than pigz, which unzips in parallel.
- Switch to downloading with aria2 instead of wget. Aria is a multiplexed downloader that can download over http and ftp. With aria2 we can establish multiple connections with the ncbi server and download much much faster.
- Eliminate needless steps from generating accessions. We are currently outputting these wgs index files that we don't use anywhere. I removed them.
- Remove needless downloads for generating accessions. We were downloading files that were not used for generating the accession index.
- ~~Use scatter gather for generating alignment indexes.~~
- Encorporate ncbi tax 2 lin into this repo and upgrade to python3

### DIFF
--- a/index-generation/Dockerfile
+++ b/index-generation/Dockerfile
@@ -27,8 +27,11 @@ RUN apt-get update && \
         python3-pip \
         software-properties-common
 
-# Install s3parcp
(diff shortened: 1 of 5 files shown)


morsecodist:
Using scatter gather to parallelize index generation results in memory issues so I reverted that element of the changes.

rzlim08 (approved):
LGTM

PR #64, opened 2022-05-19
Fall back to batch API for status checks

morsecodist:
I suspect we are hitting the lambda limit in the lambda that updates S3 with batch job information. This is causing them to not update which means we wait forever for the finished status. As a temporary fix I have added a fallback to eventually use the real API every ~30 minutes so the jobs will terminate eventually.


rzlim08 (approved):
Looks good! Thanks for doing this so quickly

PR #70, opened 2022-05-31
host_filter.wdl modernization

mlin:
copilot:summary


short-read-mngs/host_filter.wdl
```diff
@@ -1,674 +1,619 @@
...
+
+    File adapter_fasta
+
+    String host_genome
+    File bowtie2_index_tar
+    File hisat2_index_tar
+    File kallisto_idx
+
+    File human_bowtie2_index_tar
+    File human_hisat2_index_tar
```
ovalenzuela19, line 22: what if the host genome is not a human?

ovalenzuela19: Ahh this makes sense! Thanks so much!

short-read-mngs/host_filter.wdl
```diff
@@ -1,674 +1,619 @@
...
-    String docker_image_id
-    String s3_wd_uri
-    File validate_input_summary_json
-    Array[File] valid_input_fastq
-    File star_genome
-    String nucleotide_type
-    String host_genome
-    String genome_dir = "STAR_genome/part-0/"
+  # Adapter trimming and QC filtering
+  call fastp_qc {
```
ovalenzuela19, line 47: nit: can we rename the steps slightly? I believe our step names are in camelCase and start with a verb

ovalenzuela19: e.g. `fastp_qc` => `RunFastpQc`

mlin:
@ovalenzuela19 @morsecodist @rzlim08 @katrinakalantar Please have a look over this PR before merging at last. I took out the indexing from this since that's subsumed in #182. Note that this *replaces* the existing host_filter.wdl, not sure if there might be value in keeping both versions around (on HEAD)?

mlin:
@ovalenzuela19 @morsecodist @rzlim08 @katrinakalantar In today's pipeline meeting we discussed that indeed, it seems like a good idea to keep both versions of the WDL alive on the main branch, since we expect to continue supporting the original version for some time. 

One way to do this would be to keep the current diff that replaces host_filter.wdl and rewires other parts that refers to it; and simply add `original_host_filter.wdl` alongside. There are other ways of course, but let me put that one forward as a strawperson.

Thoughts? would this need some change to whatever part of the system actually launches the original version if so requested?

workflows/short-read-mngs/test/test_short_read_mngs.py
```diff
@@ -31,7 +31,7 @@ def test_bench3_viral(short_read_mngs_bench3_viral_outputs):
         taxon_counts = json.load(infile)["pipeline_output"]["taxon_counts_attributes"]
 
     taxa = set(entry["tax_id"] for entry in taxon_counts)
-    assert len(taxa) == 177
+    assert abs(len(taxa) - 184) < 16
```
rzlim08, line 34: Are we expecting to introduce some variability into the host filtering? Or was this just a holdover from the old assert?

rzlim08:
> @ovalenzuela19 @morsecodist @rzlim08 @katrinakalantar In today's pipeline meeting we discussed that indeed, it seems like a good idea to keep both versions of the WDL alive on the main branch, since we expect to continue supporting the original version for some time.
> 
> One way to do this would be to keep the current diff that replaces host_filter.wdl and rewires other parts that refers to it; and simply add `original_host_filter.wdl` alongside. There are other ways of course, but let me put that one forward as a strawperson.
> 
> Thoughts? would this need some change to whatever part of the system actually launches the original version if so requested?

Thanks Mike and sorry for the late review. 

One option is to just use `host_filter.wdl` as a workflow manager and reference other `wdl` files e.g. (`old_host_filter.wdl`, `new_host_filter.wdl`). The `wdl` files already get `miniwdl zip`ped up so we could just call the zip files from the webapp. 

I'm not sure if it's possible to have wdl expect different inputs given a flag, but that would be nice. Otherwise I guess we'd have to made almost all of the differing inputs optional.

mlin:
> I'm not sure if it's possible to have wdl expect different inputs given a flag, but that would be nice. Otherwise I guess we'd have to made almost all of the differing inputs optional.

@rzlim08 I think it would be closer to the second case unfortunately. Would you still want to see that approach in view of that potential awkwardness? (Thx- I'm just not familiar enough with the pieces of code actually invoking these workflows to understand the constraints here)

rzlim08:
> > I'm not sure if it's possible to have wdl expect different inputs given a flag, but that would be nice. Otherwise I guess we'd have to made almost all of the differing inputs optional.
> 
> @rzlim08 I think it would be closer to the second case unfortunately. Would you still want to see that approach in view of that potential awkwardness? (Thx- I'm just not familiar enough with the pieces of code actually invoking these workflows to understand the constraints here)

I think the other option would be to run with 2 WDL files and the webapp could choose between them. This would likely mean one of the wdl's would be set as a "default" for e.g. local testing/ benchmarking, but the webapp could have the logic to run one vs the other. This might be the easiest solution for now.

mlin:
@ovalenzuela19 The tentative solution here is to merge with the modern host filtering WDL *replacing* the original one as `host_filter.wdl`, and adding the original one alongside as `original_host_filter.wdl`. Everything else in the repo (`local_driver.wdl`, `amr/run.wdl`, unit tests, etc.) would be wired to `host_filter.wdl` and not `original_host_filter.wdl`.

Question for you- is that going to work for the webapp as far as how it invokes the original pipeline when asked to do so? Or does it need the zip to have `host_filter.wdl` as the original pipeline, in which case there would not be much point (and might cause additional confusion) to add `original_host_filter.wdl` at all.

Thanks, we're obviously trying to avoid breaking anything that depends on the structure of this repo that aren't necessarily obvious when making these kinds of changes.

ovalenzuela19:
@mlin so would `original_host_filter.wdl` just be there for reference and not really usable?

I think from a webapp perspective that solution should be fine

mlin:
@ovalenzuela19

> would original_host_filter.wdl just be there for reference and not really usable?

Mostly yes, for reference. The unanswered bit is what would we do if a user absolutely needs us to make some further change/bugfix to the *original* pipeline they're still depending on. How would we roll that out as a tagged release & zip file the webapp can use, once it's been renamed to `original_host_filter.wdl`? 

If we're optimists we would say that's fairly unlikely to happen, to the extent we can punt now and figure it out later if it does, but, thought we should at least air it out here. cc @rzlim08 @morsecodist @katrinakalantar for visibility

rzlim08:
> @ovalenzuela19
> 
> > would original_host_filter.wdl just be there for reference and not really usable?
> 
> Mostly yes, for reference. The unanswered bit is what would we do if a user absolutely needs us to make some further change/bugfix to the _original_ pipeline they're still depending on. How would we roll that out as a tagged release & zip file the webapp can use, once it's been renamed to `original_host_filter.wdl`?
> 
> If we're optimists we would say that's fairly unlikely to happen, to the extent we can punt now and figure it out later if it does, but, thought we should at least air it out here. cc @rzlim08 @morsecodist @katrinakalantar for visibility

Yeah I think the main question is how to support both pipelines at the same time indefinitely. If we have both WDLs we'd at least be able to patch either. I'll see if I can make a small customization to pin a release to a major/minor version. That being said I don't think we should be running with this forever and would like to drop support for the old version eventually if we're comfortable enough with the new one.

ovalenzuela19:
> Yeah I think the main question is how to support both pipelines at the same time indefinitely. If we have both WDLs we'd at least be able to patch either. I'll see if I can make a small customization to pin a release to a major/minor version. That being said I don't think we should be running with this forever and would like to drop support for the old version eventually if we're comfortable enough with the new one.

What if we just create a new workflow that is just the old short-read-mngs pipeline with the old host filtering wdl? That way you can still package it like normal and it's up to the web app which wdl will be used?

kislyuk:
Whoa. Just noticed this! Huge step forward for czid. Congrats on landing this!

PR #71, opened 2022-06-03
Index generation load patch

morsecodist:
Contains a few patches for issues I discovered from the first run on a full database:
- Fix an issue where I was creating 1 row per taxon per level instead of 1 row per taxon
- Fix an issue where pre-existing taxons end up without version info
- Update some details about the mysql import
- Add alignment config update


rzlim08 (approved):
LGTM

PR #77, opened 2022-06-14
 add longest 5 reads for nr

morsecodist:
We have these outputs for the blast feature for NT, this adds the same for NR. I did a small refactor to allow for code reuse. In researching this addition I did notice quite a few bits of weirdness with experimental, for example this step is receiving a lot of outputs it isn't using, and I think we are producing outputs we never use. I also think the long term home for this may be somewhere else, as we discussed. All that said I think the way I did this gives us the feature without incurring additional tech debt. I refactored the step to use the raw input file instead of operating on the alignment viz files. The top n code is now a lot simpler and all in one place.

Also added another assertion in the tests.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/steps/generate_alignment_viz.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/steps/generate_alignment_viz.py
@@ -6,7 +6,6 @@ import traceback
 from collections import defaultdict
 import subprocess
 import threading
-from typing import Dict, Union
 
 from Bio.Seq import Seq
 from Bio.SeqRecord import SeqRecord
--- a/short-read-mngs/idseq-dag/idseq_dag/steps/generate_alignment_viz.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/steps/generate_alignment_viz.py
@@ -103,7 +103,6 @@ class PipelineStepGenerateAlignmentViz(PipelineStep):
         with open(summary_file_name, 'w') as summary_f:
             summary_f.write(summary_msg)
 
-
     @staticmethod
     def output_n_longest_reads(db_type: str, annotated_fasta: str, output_longest_reads_dir: str, n=5):
         n_longest = {}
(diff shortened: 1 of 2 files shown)


short-read-mngs/idseq-dag/idseq_dag/steps/generate_alignment_viz.py
```diff
@@ -88,9 +88,13 @@ def safe_multi_delete(files):
...
         deleter_thread.start()
 
-        self.dump_align_viz_json(output_json_dir, output_longest_reads_dir, db_type, result_dict)
+        self.dump_align_viz_json(output_json_dir, db_type, result_dict)
 
         deleter_thread.join()
+        read2seq = {}  # to free up memory
+        command.make_dirs(output_longest_reads_dir)
+        PipelineStepGenerateAlignmentViz.output_n_longest_reads("nt", annotated_nt_fasta, output_longest_reads_dir)
+        PipelineStepGenerateAlignmentViz.output_n_longest_reads("nr", annotated_nr_fasta, output_longest_reads_dir)
```
ovalenzuela19, line 97: love the refactor!

short-read-mngs/idseq-dag/idseq_dag/steps/generate_alignment_viz.py
```diff
@@ -42,14 +41,15 @@ def run(self):
         db_type = "nt"  # Only NT supported for now
         # TODO: Design a way to map in/out files more robustly, e.g. by name/type
         annotated_m8 = self.input_files_local[0][0]
-        annotated_fasta = self.input_files_local[1][0]
+        annotated_nt_fasta = self.input_files_local[1][0]
+        annotated_nr_fasta = self.input_files_local[1][2]
```
ovalenzuela19, line 45: I'm not familiar with the structure of `input_files_local` is.Is the structure defined somewhere?

ovalenzuela19 (approved):
LGTM! 🚀

short-read-mngs/idseq-dag/idseq_dag/steps/generate_alignment_viz.py
morsecodist: It is a holdover from idseq-dag unfortunately

PR #88, opened 2022-07-15
Use marisa trie in CG

morsecodist:


consensus-genome/test/test_wdl.py
```diff
@@ -142,31 +153,33 @@ def test_sars_cov2_ont_cg_input_file_format(self):
...
-            self.assertTrue(fh.readline().startswith("AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTG"))
-
-        with self.assertRaises(CalledProcessError) as ecm:
-            self.run_miniwdl(task="FetchSequenceByAccessionId", args=["accession_id=NO_ACCESSION_ID"])
-        self.assertRunFailed(ecm, task="FetchSequenceByAccessionId",
-                             error="AccessionIdNotFound",
-                             cause=("The Accession ID was not found in the CZ ID database, "
-                                    "so a generalized consensus genome could not be run"))
-
-    def test_fetch_sequence_by_expired_accession_id(self):
```
morsecodist, line 162: I removed this test because NR/NT should have expired versions so the increment logic shouldn't be needed anymore.

rzlim08 (approved):
This looks great!

PR #93, opened 2022-07-22
Use Marisa Trie in short-read-mngs

morsecodist:


rzlim08 (approved):
This looks good. Do you think there's a way of storing the type eg. `QII` `lll` with the file? It's not a blocker or anything but it would be nice to infer that

morsecodist:
@rzlim08 I agree needing to know the type is a pain. I considered ways around it it but the issue is I want to use the mmap IO. You can pickle marisa-tries but then you can't use the mmap IO api. This leaves me with two options (as far as I am aware):

1. Somehow define my own binary format and make it work with marisa-trie
2. Use separate files

The first seems too complex, the second seems like I am just kind of moving the problem. 

Any other suggestions/thoughts?

PR #94, opened 2022-07-22
Add Run Glue Job 

rzlim08:
* Create a stage guarded by an input to run a glue job which calculates the zscore for the pipeline run using the default background id
* Also a small script to watch the glue job as it processes the pipeline run
* I'm still not fully sure what to do about how to handle failures. I should probably be raising more informative errors

### DIFF
--- a/short-read-mngs/experimental.wdl
+++ b/short-read-mngs/experimental.wdl
@@ -230,11 +230,9 @@ task RunGlueJob {
   set -euxo pipefail 
   BUCKET=$(echo "~{s3_wd_uri}" | cut -d/ -f 3)
   PIPELINE_RUN_ID=$(echo "~{s3_wd_uri}" | cut -d/ -f 7)
-  echo $BUCKET
   if [[ $BUCKET == "idseq-samples-sandbox" ]]; then
    ENV=sandbox
   elif [[ $BUCKET == "idseq-samples-staging" ]]; then
-   echo here
    ENV=staging
   elif [[ $BUCKET == "idseq-prod-samples-us-west-2" ]]; then
    ENV=prod
@@ -243,6 +241,7 @@ task RunGlueJob {
   fi
   aws glue start-job-run --job-name "$ENV"_heatmap_es_job --region us-west-2 --arguments "--user_pipeline_run_ids=$PIPELINE_RUN_ID --user_background_id=26 --job_type=selected_runs" 
   >>>
+
   runtime { 
     docker: docker_image_id
   }
--- a/short-read-mngs/experimental.wdl
+++ b/short-read-mngs/experimental.wdl
@@ -239,7 +239,25 @@ task RunGlueJob {
   else 
     exit 0
   fi
-  aws glue start-job-run --job-name "$ENV"_heatmap_es_job --region us-west-2 --arguments "--user_pipeline_run_ids=$PIPELINE_RUN_ID --user_background_id=26 --job_type=selected_runs" 
+  JOBID=$(aws glue start-job-run --job-name "$ENV"_heatmap_es_job --region us-west-2 --arguments="--user_pipeline_run_ids=$PIPELINE_RUN_ID,--user_background_id=26,--job_type=selected_runs" | jq .JobRunId) 
+  i=0
+  while true; do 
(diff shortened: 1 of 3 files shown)


morsecodist:
I would disable this by default via an input to preserve running outside of our infra

rzlim08:
> I would disable this by default via an input to preserve running outside of our infra

Yeah it should disable if the bucket `s3_wd_uri` isn't one of the 3 specified but I can add a default input as well

short-read-mngs/experimental.wdl
```diff
@@ -219,6 +219,35 @@ task NonhostFastq {
...
+  # This command is specific to the CZ ID system
+  
+  set -euxo pipefail 
+  BUCKET=$(echo "~{s3_wd_uri}" | cut -d/ -f 3)
+  PIPELINE_RUN_ID=$(echo "~{s3_wd_uri}" | cut -d/ -f 7)
+  echo $BUCKET
+  if [[ $BUCKET == "idseq-samples-sandbox" ]]; then
+   ENV=sandbox
+  elif [[ $BUCKET == "idseq-samples-staging" ]]; then
+   echo here
```
rzlim08, line 237: rm

short-read-mngs/experimental.wdl
```diff
@@ -219,6 +219,35 @@ task NonhostFastq {
...
+    String docker_image_id
+    String s3_wd_uri
+  }
+  command<<<
+  # This command is specific to the CZ ID system
+  
+  set -euxo pipefail 
+  BUCKET=$(echo "~{s3_wd_uri}" | cut -d/ -f 3)
+  PIPELINE_RUN_ID=$(echo "~{s3_wd_uri}" | cut -d/ -f 7)
+  echo $BUCKET
```
rzlim08, line 233: rm

short-read-mngs/experimental.wdl
```diff
@@ -219,6 +219,52 @@ task NonhostFastq {
...
+  else 
+    exit 0
+  fi
+  JOBID=$(aws glue start-job-run --job-name "$ENV"_heatmap_es_job --region us-west-2 --arguments="--user_pipeline_run_ids=$PIPELINE_RUN_ID,--user_background_id=26,--job_type=selected_runs" | jq -r .JobRunId) 
+  i=0
+  while true; do 
+    JOBSTATE=$(aws glue get-job-run --job-name staging_heatmap_es_job --run-id $JOBID | jq -r .JobRun.JobRunState)
+    if [[ $JOBSTATE == "SUCCEEDED" ]]; then 
+      exit 0 
+    elif [[ $JOBSTATE == "FAILED" ]]; then
```
nigel-smk, line 248: I think that you will have to handle more potential states. https://docs.aws.amazon.com/glue/latest/dg/job-run-statuses.html

nigel-smk (changes requested):

short-read-mngs/experimental.wdl
```diff
@@ -247,6 +247,10 @@ task RunGlueJob {
       exit 0 
     elif [[ $JOBSTATE == "FAILED" ]]; then 
       exit 1
+    elif [[ $JOBSTATE == "ERROR" ]]; then 
+      exit 1
+    elif [[ $JOBSTATE == "TIMEOUT" ]]; then 
+      exit 1
     else 
       sleep 60
```
nigel-smk, line 255: I think STOPPED should be added as a case at least in which we exit. I guess it would exit on the timeout. That would be an edge case in which the job gets manually stopped via CLI or console

nigel-smk (changes requested):

short-read-mngs/experimental.wdl
rzlim08: Hm, good point I wasn't sure if STOPPED would be an error state or a success state. Or if it would eventually become one. Do you know if there is any chance that a STOPPED state would eventually lead to SUCCESSFUL?

short-read-mngs/experimental.wdl
rzlim08: Thanks!

morsecodist (approved):

PR #97, opened 2022-08-10
Revert fetching accessions dynamically from S3

morsecodist:
Thanks to @rzlim08 for catching this!

When I did the maris-trie refactor I also switched to only downloading the required accessions from NT and NR instead of pre-downloading all of them. This could be faster if it was implemented with parallelism and a global cache but it was not so it turned out to be way slower. We may want to do this optimization properly to save space as NT and NR grow in the future but for now I am just reverting.

Reverts some of https://github.com/chanzuckerberg/czid-workflows/pull/93

The existing tests should ensure this works, outputs should be identical.

### DIFF
--- a/short-read-mngs/idseq-dag/idseq_dag/steps/download_accessions.py
+++ b/short-read-mngs/idseq-dag/idseq_dag/steps/download_accessions.py
@@ -1,16 +1,9 @@
 import re
-from urllib.parse import urlparse
 from idseq_dag.engine.pipeline_step import PipelineStep
 import idseq_dag.util.s3 as s3
 import idseq_dag.util.m8 as m8
 
 from idseq_dag.util.dict import open_file_db_by_extension
-
-import boto3
-import botocore.exceptions
-from botocore import UNSIGNED
-from botocore.client import Config
-
 
 MAX_ACCESSION_SEQUENCE_LEN = 100000000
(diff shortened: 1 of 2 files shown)


short-read-mngs/idseq-dag/idseq_dag/steps/download_accessions.py
```diff
@@ -31,7 +24,12 @@ def run(self):
         # db_type = self.additional_attributes["db_type"]
         (_read_dict, accession_dict, _selected_genera) = m8.summarize_hits(hit_summary)
         with open_file_db_by_extension(loc_db, "QII", stringify=False) as loc_dict:
-            self.download_ref_sequences_from_file(accession_dict, loc_dict, db_s3_path, output_reference_fasta)
+            db_path = s3.fetch_reference(
```
rzlim08, line 27: since you changed the db type to File won't this automatically get downloaded by miniwdl?

morsecodist: Yeah iirc the `fetch_reference` code detects that and won't re-download. I think I can technically remove this but I just did a full revert for simplicity's sake.

morsecodist: Yeah just confirmed this is the case.

rzlim08 (approved):

PR #100, opened 2022-08-16
Refactor AMR wdl

rzlim08:
* Gets rid of conda in dockerfile
* Adds the databases into the dockerfile so we don't have to download/load them dynamically
* Clean up a lot of the python code


amr/Dockerfile
```diff
@@ -2,24 +2,29 @@ FROM ubuntu:20.04
...
-RUN apt-get update && apt-get -y install curl git python3 python3-pip
+RUN ln -s /usr/bin/python3 /usr/bin/python
 
-RUN pip3 install pandas
+RUN curl -L https://github.com/bbuchfink/diamond/releases/download/v0.8.36/diamond-linux64.tar.gz | tar xz && mv diamond /usr/local/bin
+RUN curl -L http://www.unafold.org/download/oligoarrayaux-3.8.tar.gz | tar xz && cd oligoarrayaux-3.8/ && ./configure && make install 
+RUN curl -L https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2 | tar xj && cd samtools-1.9/ && make && mv samtools /usr/local/bin/
+RUN curl -L https://github.com/arq5x/bedtools2/releases/download/v2.29.1/bedtools-2.29.1.tar.gz | tar xz && cd bedtools2 && make && mv bin/* /usr/local/bin/
+RUN  curl -L https://github.com/lh3/bwa/releases/download/v0.7.17/bwa-0.7.17.tar.bz2 | tar xj && cd bwa-0.7.17/ && make &&  mv bwa /usr/local/bin/
+RUN curl -L  https://bitbucket.org/genomicepidemiology/kma/get/1.4.3.tar.gz | tar xz && cd genomicepidemiology-kma-47505a7dd4eb/ && make && mv kma kma_index kma_shm kma_update /usr/local/bin/
```
jfoo1984, line 14: This might be ignorance on my part, but it feels like this is a varied collection of different tools.  Would it be helpful for future reference (in case updates have to be made or to resolve issues) to have comments explaining what each of these are for?

rzlim08: This is a good question, to be honest, most of these are just dependencies of `rgi`, which I just copied from the conda to docker. I don't think I'd be able to say specifically what each one is used for, other than very broadly (e.g. bedtools is used to manipulated `.bed` files. etc)

jfoo1984: Ah, thanks for the clarification!  Maybe just a single comment indicating these are dependencies of `rgi` would be helpful, and enough of a hint for someone to get context later on?

jfoo1984 (approved):

amr/Dockerfile
rzlim08: Thanks!

PR #103, opened 2022-08-19
add contig name to output of coverage viz

rzlim08:


rzlim08:
Closing until this is on the docket

PR #106, opened 2022-08-30
reduce alignment scalability memory usage 

rzlim08:
* Reduces memory for batch alignment scalability instances 
* Increases max chunks in flight to 30

### DIFF
--- a/short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
+++ b/short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
@@ -21,7 +21,7 @@ from botocore.config import Config
 
 log = logging.getLogger(__name__)
 
-MAX_CHUNKS_IN_FLIGHT = 10
+MAX_CHUNKS_IN_FLIGHT = 30
 ALIGNMENT_WDL_VERSIONS: Dict[str, str] = {
     "diamond": "v1.0.0",
     "minimap2": "v1.0.0",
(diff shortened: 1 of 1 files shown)


short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -21,7 +21,7 @@
 
 log = logging.getLogger(__name__)
 
-MAX_CHUNKS_IN_FLIGHT = 10
+MAX_CHUNKS_IN_FLIGHT = 30
```
morsecodist, line 24: I actually forgot about this...

I actually think we should not have a max chunks in flight, and having it is kind of a disaster. We won't submit all the chunks at once we will wait for our first chunks to complete before submitting later chunks. However, in the meantime new chunks can enter the queue so we will now have the later chunks waiting on all of them keeping the alignment job running. This makes scaling way worse and if we hit timeouts and get even more retries on top of that we can have an explosion in waiting time. The max chunks in flight really doesn't protect us from going to too high of a scale because we can get too many alignment jobs running, it really only makes things worse.

I think making it 30 does this in effect because we don't have more than 30 chunks for any of our indexes. Would you mind adding a comment to point that out and we can re-visit removing this completely in the future?

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -97,10 +97,10 @@ def _run_batch_job(
         jobDefinition=job_definition,
         containerOverrides={
             "environment": [{"name": k, "value": v} for k, v in environment.items()],
-            # (524288 - 1024) / 2, 524288 = r5d.24xlarge memory
-            #   2 = jobs per instance, 1024 = leftover for other processes
-            "memory": 261632,
-            "vcpus": 48,  # 96 / 2, 96 = r5d.24xlarge vcpus , 2 = jobs per instance
+            # (524288 - 1024) / 4, 524288 = r5d.24xlarge memory
```
morsecodist, line 100: Do we want to just remove this math now that we are likely moving to other instance types soon? This was taken directly from the old scaling math which was actually based on gsnap/rapsearch CPU saturation. I think this may be irrelevant in this brave new world.

morsecodist (approved):
Code looks good just a few comments about comments.

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
rzlim08: Removed, I also think the amount of memory on the r5d.24X instance was wrong

short-read-mngs/idseq_utils/idseq_utils/batch_run_helpers.py
rzlim08: Done, yeah the intention was to effectively remove this bottleneck

PR #108, opened 2022-09-07
Add reference to host filtering

rzlim08:
* Adds ability to run host filtering on input files
* Requires that we use inputs for host filtering regardless of whether host filtering runs
* adds RunSpades task

### DIFF
--- a/amr/run.wdl
+++ b/amr/run.wdl
@@ -7,7 +7,7 @@ workflow amr {
         Array[File]? non_host_reads
         File? raw_reads_0
         File? raw_reads_1
-        File contigs
+        File? contigs
         String docker_image_id
         String host_filtering_docker_image_id = "czid-short-read-mngs" # default local value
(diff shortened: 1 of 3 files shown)


amr/run.wdl
```diff
@@ -1,21 +1,40 @@
...
         # Dummy values - required by SFN interface
         String s3_wd_uri = ""
     }
-
+    if (defined(raw_reads_0)) { 
+        call host_filter.czid_host_filter as host_filter_stage { 
+            input:
+            fastqs_0 = select_first([raw_reads_0]),
+            fastqs_1 = if defined(raw_reads_1) then select_first([raw_reads_1]) else None,
+            docker_image_id = host_filtering_docker_image_id,
```
rzlim08, line 25: In this case we are using different docker images for host filtering compared to the AMR pipeline. If this causes a problem we can merge these into a large image

morsecodist (approved):

PR #112, opened 2022-09-16
Add reduplicate stage to host filtering

rzlim08:
* takes the outputs from czid-dedup and outputs the duplicated reads (not including the original reads) 
* appending the dedup plus redup files should create the un-deduplicated output file


morsecodist:
The code seems fine but why is this needed?

rzlim08:
Closing. Need to do this, but for the new host filtering

PR #115, opened 2022-09-27
Fix phylotree missing accessions

rzlim08:
* This was tricky to see before running it, when phylotrees don't get an accession, they defaulted to an empty string 
* Also added a test, and ran `black` on phylotree test_wdl.py


morsecodist (approved):
Thank you!

PR #116, opened 2022-09-27
Increment consensus-genome accession if not found

rzlim08:
* increments the version of the accession by 1 if it's not found in the database
* also ran `black` on this test_wdl.py


morsecodist (approved):
Thank you!

PR #120, opened 2022-09-30
cancel batch job if parent instance is killed

rzlim08:
* not sure if this will work, but just a thought. Having the batch jobs sit in the queue even after the underlying job has been cancelled or killed seems like it might be a pretty big bottleneck
* Unsure if the `finally` block will trigger on SPOT interruptions


morsecodist (approved):
Great idea!

PR #123, opened 2022-10-03
refactor the full benchmark trigger to use czid name

morsecodist:
To be paired with https://github.com/chanzuckerberg/idseq/pull/446. Refactors the github actions input names to use czid instead of idseq.


rzlim08 (approved):
Looks good, thanks!

PR #124, opened 2022-10-04
run benchmarks simultaneously

morsecodist:
After further discussions with @mlin I think I finally got to the bottom of this. There was the element of controlling time tracking that he mentioned, but that didn't explain why the lock was called `_timestamp_lock` and why there was a 1.1 second wait time. 

What does explain it is that `run_sfn` uses second-precision timestamps as unique identifiers. This means you can't invoke `run_sfn` twice in the same second or you will get a name conflict. This explains the `_timestamp_lock` (it is a lock to guarantee unique timestamps) and the 1.1 second wait time (needs to be longer than a second to guarantee it will never be invoked within the same second.

This means this change actually is necessary now that we want to move towards concurrent runs for our benchmarks. I updated the comments with a detailed description of why this behavior is necessary and how it works and this is ready for review!

## Previous Description (preserved for historical context)

I am raising this as a draft PR to communicate what I am thinking here. This actually might not be necessary. From the looks of the code here (like the 1.1 second sleep) it appears the intent may have been to prevent `run_sfn` from being invoked at the same time by multiple threads. Alternatively, the intent really could have been to sequence them one after another. This seems odd to me because a lot of effort was put into parallelizing these in the first place, hence the source of confusion.

If the issue is preventing `run_sfn` from being invoked at the same time this will solve it. It starts the process, then releases the lock, then waits for the process to complete. All of this is happening inside of a future which handles the parallelism.

If the intent was running one at a time then we should close this PR and discuss why. If we want to keep it that way we can remove the futures stuff.

### DIFF
--- a/short-read-mngs/auto_benchmark/run_dev.py
+++ b/short-read-mngs/auto_benchmark/run_dev.py
@@ -170,7 +170,7 @@ def run_sample(czid_repo, workflow_version, index_version, settings, key_prefix,
 
     # now the lock is released
     # this is a simplified version of some of the functionality of subprocess.run
-    
+
     # wait for the process to complete and get stdout and stderr
     stdout, stderr = process.communicate()
     # get the return code of the process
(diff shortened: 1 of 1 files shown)


mlin:
@morsecodist As best I recall, the reason for running them one-at-a-time was to take out at least one confounder from the recorded wall-clock turnaround time of running each sample. However I think we can be sure there are still a number of confounders to those measurements, so I think it's a reasonable suggestion that we're better served by reducing the overall time to run all the benchmarks.

PR #126, opened 2022-10-07
Long reads ummapped function

rzlim08:
* reads in mapping from reads to contigs and hitsummaries from nt and nr
* uses python set operations to get the unmapped reads
* uses `seqkit grep` to output unmapped_reads fastq from original host filtered and subsampled input
* Added test for output and line count


morsecodist (changes requested):
The change looks good but can you add some test assertions? At least for the presence of the file.

long-read-mngs/run.wdl
```diff
@@ -455,18 +455,59 @@ task TallyHits {
...
+
+        unmapped_reads = set()
+        with open("all_reads.txt", "r") as ar:
+            for line in ar:
+                read, contig = line.strip().split("\t")
+                if read not in mapped_hits and contig not in mapped_hits:
+                    unmapped_reads.add(read)
+
+        with open("unmapped_reads.txt", "w") as ur:
+            for read in unmapped_reads:
```
morsecodist, line 495: I think this can just be:

```python
ur.writelines(unmapped_reads)
```

morsecodist (approved):

long-read-mngs/run.wdl
rzlim08: I think it doesn't output with the newlines, possibly because it's a set?

PR #132, opened 2022-10-24
fix issue with e value calculation

morsecodist:
We were getting overflow errors because of samples with e-values that were way too high. This PR effectively adds a maximum to e-values so this no longer happens.


rzlim08 (approved):
Thanks! Should we change the short-read-mngs file as well?

morsecodist:
@rzlim08 I will be merging them very soon and I will use this version post-merge

PR #134, opened 2022-10-25
Eliminate downloading all of NR and NT

morsecodist:
Avoid downloading all of NR and NT by downloading chunks using the indexes we already have and range queries. In order to mitigate the potential slowdown this might cause I wrote the parallel download in Go and I am calling it using python bindings. Go makes it easy to get good concurrent IO performance, it is kind of ideal for this use case. The Go code is very minimal.

We were downloading all of NR and NT for two things. The first was creating fastas from all of the aligned accessions. For this I created a go script that outputs directly to a file using `WriteAt`. This keeps memory usage very low because this file can get large. It also makes the code cleaner in my opinion.

The second is for the alignment viz. We had a file based code path here but I actually removed it because this was only used way back when we ran phylo tree generation on the web app containers. We always just load the sequences into memory now. So I modified that to read the sequences from S3 into memory.


short-read-mngs/Dockerfile
```diff
@@ -1,3 +1,27 @@
...
+RUN tar -C . -xzf go${GO_VERSION}.linux-amd64.tar.gz && rm go${GO_VERSION}.linux-amd64.tar.gz
+ENV GOPATH /go
+ENV PATH $GOPATH/bin:$PATH
+RUN /go/bin/go version
+WORKDIR /go/src/s3quilt
+COPY s3quilt/go.mod s3quilt/go.sum s3quilt/s3quilt.go ./
+
+RUN go get golang.org/x/tools/cmd/goimports
+RUN go install golang.org/x/tools/cmd/goimports
+RUN go get github.com/go-python/gopy
```
jgadling, line 16: looks like these aren't versioned. Might want to look into tacking in `go mod` support at some point

jgadling (approved):

short-read-mngs/Dockerfile
morsecodist: fixed!

PR #136, opened 2022-11-02
boto3 adaptive retries

morsecodist:
Originally this ticket was to add an exponential backoff. But it turns out we already have it and we were still hitting the limit. So I took the opportunity to switch to the `adaptive` retry mode as suggested by @rzlim08. This should proactively prevent hitting the rate limit by throttling requests on the client. Doing this will be helpful for our use case because we kick off a lot of jobs in parallel in the exact same moment. This throttling will add some wait time before making all of those requests reducing the odds that we hit the rate limit. Using this requires a new version of boto3. I also increased the max attempts.

While dealing with upgrading the boto3 versioning I ran into some conflicts with the AWS cli version and realized there is no reason why we should still need to install this for idseq-dag now that we interact with AWS services almost exclusively through the SDK. I was able to clean up a lot of legacy code that was depending on the AWS CLI that hasn't been used in well over a year and never will be used again and removed our dependency on the now deprecated V1 of the AWS CLI.


workflows/short-read-mngs/Dockerfile
```diff
@@ -75,8 +75,7 @@ RUN apt-get -q install -y \
         spades \
         ncbi-blast+
 
-# Match versions from Ubuntu 20.04 LTS (please do not upgrade except to fix bugs)
```
morsecodist, line 78: I don't see a great reason to pin these to the 20.04 LTS

lib/idseq-dag/idseq_dag/steps/prepare_taxon_fasta.py
```diff
@@ -25,7 +26,7 @@ def run(self):
             local_basename = f"{hit_type}_{os.path.basename(output_file)}.fasta"
             bucket, key = s3.split_identifiers(s3_file)
             local_file = os.path.join(self.output_dir_local, local_basename)
-            s3.fetch_byterange(first_byte, last_byte, bucket, key, local_file)
+            download_chunks_to_file(bucket, key, local_file, [first_byte], [last_byte - first_byte])
```
morsecodist, line 29: I just added this new function for another use case so we can just use it here instead of relying on the old function which used the AWS CLI.

rzlim08 (approved):
This looks good, thanks!

PR #141, opened 2022-11-09
increase block size for long reads pipeline

rzlim08:
The same error as in https://github.com/chanzuckerberg/czid-workflows/pull/39 started occurring for the long-read pipeline randomly. 

I've increased the block-size by 10X but am unsure if this will account for extremely large samples.


morsecodist (approved):
Can you actually merge to main instead, sorry for the reversal

PR #146, opened 2022-11-19
Disable coverage viz

rzlim08:


morsecodist:
Coverage viz is now working

PR #152, opened 2022-12-01
output sequence length rather than sequence

rzlim08:
* The python csv module was breaking because we were outputting the sequence and it was too long
* Use `awk` to just output the sequence length

### DIFF
--- a/workflows/long-read-mngs/run.wdl
+++ b/workflows/long-read-mngs/run.wdl
@@ -308,7 +308,7 @@ task RunReadsToContigs {
 
         # convert non-contigs.fastq file to .fasta file
         seqtk seq -a sample.non_contigs.fastq > sample.non_contigs.fasta
-        samtools view -F 0x900 sample.reads_to_contigs.sam | grep -v "^@" | awk '{ print $1 "\t" $3 "\t" length($10)}' > reads_to_contigs.tsv
+        samtools view -F 0x900 sample.reads_to_contigs.sam | awk '{ print $1 "\t" $3 "\t" length($10)}' > reads_to_contigs.tsv
     >>>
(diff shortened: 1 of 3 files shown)


workflows/long-read-mngs/run.wdl
```diff
@@ -308,7 +308,7 @@ task RunReadsToContigs {
 
         # convert non-contigs.fastq file to .fasta file
         seqtk seq -a sample.non_contigs.fastq > sample.non_contigs.fasta
-        samtools view -F 0x900 sample.reads_to_contigs.sam | grep -v "^@" | cut -f1,3,10 > reads_to_contigs.tsv
+        samtools view -F 0x900 sample.reads_to_contigs.sam | grep -v "^@" | awk '{ print $1 "\t" $3 "\t" length($10)}' > reads_to_contigs.tsv
```
robertaboukhalil, line 311: Curious: are you using `grep -v` to remove the SAM header lines? If so, one alternative is to use the `--no-header` flag: `samtools view --no-header -F 0x900 sample.reads_to_contigs.sam`

robertaboukhalil (approved):
One minor suggestion, otherwise looks good to me!

workflows/long-read-mngs/run.wdl
robertaboukhalil: Oh actually, according to [the docs](http://www.htslib.org/doc/samtools-view.html#OPTIONS), `--no-header` is the default, so you might be able to get away with just remove grep

rzlim08: thanks! I think this was a leftover from when were just `cat`-ing the output.

PR #154, opened 2022-12-02
add minimap2 to host genome generation

morsecodist:
long-read-mngs requires these minimap2 host indexes so I am adding them here.


rzlim08 (approved):
Thanks!

PR #155, opened 2022-12-05
optimize memory usage of alignment viz

morsecodist:
When switching to `s3quilt` I modified the program to load all of the sequence data into memory, then add it to `accession2seq`. At the time I thought it would be fine because I assumed that `accession2seq` would just get a pointer to the string in `sequences` but in the loop we actually create a substring which effectively double stores all of the sequences in memory. I modified the approach to only download 500 sequences at a time, then write those sequences to `accession2seq`, then download another chunk, releasing the downloaded chunk. This should keep memory usage pretty close to what it was before the switch while still benefiting from the parallel downloads.  `sequences` is also storing the accession name while `accession2seq` doesn't need that. Fortunately, we have an index for both the header (accession name) length and the sequence length so I can modify my start and length values to start at the end of the header and use just the sequence length so we no longer need to download the accession names at all. This probably isn't a huge deal because the names are way smaller than the sequences but it was an easy win.

### DIFF
--- a/workflows/short-read-mngs/experimental.wdl
+++ b/workflows/short-read-mngs/experimental.wdl
@@ -225,6 +225,7 @@ task RunGlueJob {
     String docker_image_id
     String s3_wd_uri
   }
+
   command<<<
   # This command is specific to the CZ ID system
   set -euxo pipefail 
--- a/workflows/short-read-mngs/experimental.wdl
+++ b/workflows/short-read-mngs/experimental.wdl
@@ -230,7 +230,6 @@ task RunGlueJob {
   # This command is specific to the CZ ID system
   set -euxo pipefail 
 
-
   BUCKET=$(echo "~{s3_wd_uri}" | cut -d/ -f 3)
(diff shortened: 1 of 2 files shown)


workflows/short-read-mngs/experimental.wdl
```diff
@@ -219,11 +219,13 @@ task NonhostFastq {
     docker: docker_image_id
   }
 }
+
 task RunGlueJob {
   input {
     String docker_image_id
     String s3_wd_uri
   }
+
```
morsecodist, line 228: I had to make changes in `workflows/short-read-mngs` to trigger CICD :/ I should probably fix this at some point

morsecodist:
Oh damn this is a go slice not a list so I can't pop it. I either need to go back to writing it as a file, do some kind of chunking, or see if I can convert to a list while just using pointers to strings. I will do a quick experiment to see if `list` results in double storage and if not I will write to a file.

morsecodist:
The benchmarks are failing with this weird intermittent sklearn issue that is not effected by this. I think this is safe to merge

rzlim08 (approved):
Looks good!

PR #160, opened 2022-12-14
Fix contig base counts

rzlim08:
We are deduplicating chimeric reads from the sam file for the reads_to_bases csv but not for the sam file itself. This prevents the chimeric reads from being doubly counted.


scripts/release.sh
```diff
@@ -29,7 +29,8 @@ else
 fi
 
 if [[ $( git branch --show-current) != "main" ]]; then 
-    TAG=$TAG"-beta"
+    COMMIT=$(git rev-parse --short HEAD)
+    TAG=$TAG"-$COMMIT"
```
rzlim08, line 33: changes the name of a beta tag to the commit id to get rid of collisions when working of simultaneous branches

PR #161, opened 2022-12-16
missed a place where min_contig_size is used

rzlim08:
* If we don't set this, it defaults to 4


rzlim08:
@morsecodist merging this one so we don't lose it

PR #165, opened 2023-01-17
remove contigs with no mapped reads in long-read-mngs

morsecodist:
As discussed we were having problems in the application because some contigs had no reads mapped to them when generating our reads to contigs mapping. This distorts some of our metrics, producing nonsense values. @katrinakalantar determined that likely nothing would be lost by just removing these contigs, so that is what this PR does.


katrinakalantar (commented):
This looks good to me!

PR #168, opened 2023-01-19
small postprocessing discrepancies with long-read-posprocessing pr

rzlim08:
* fixes some small discrepancies with the post-processing step that came in from long-read-mngs


morsecodist (approved):
Thanks for catching this!

PR #171, opened 2023-01-20
sort bowtie2 outputs for repeatablity

rzlim08:
* Looking at repeatability of the short-read-mngs pipeline
* For the github actions test, it looks like the main reason that outputs are different are just because bowtie2 outputs a SAM that's randomly sorted. Sorting this should allow us to get the same results repeatedly. 
* Repeatability might not be possible for more complex samples. We can't seed bowtie2 for tiebreakers since we run it using multiple cores, but at least for our simple test samples this seems to work
* Some files that don't end up being repeatable are: STAR and SRST log files, since they have dates in them, the "Merged" taxon counts, which don't get used, and SAM files, since they print out the number of cores that a machine is using. 
* Otherwise, it looks like running the same file twice should get the exact same results now

### DIFF
--- a/.github/workflows/wdl-ci-integration.yml
+++ b/.github/workflows/wdl-ci-integration.yml
@@ -26,11 +26,12 @@ jobs:
           fetch-depth: 2
       - id: list_dirs
         # only run on wdl dirs with diff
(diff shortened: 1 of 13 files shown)


morsecodist (approved):
Looks good once the testing issues are resolved.

PR #175, opened 2023-01-30
[modernized host filter] add ERCC and gene-level outputs to kallisto

mlin:
The kallisto step gains two new derivative output files:
* `ERCC_counts.tsv`: Estimated read counts for the ERCC sequences only (two-column TSV: ERCC_id, est_counts)
* `gene_abundance.tsv`: gene-level est_counts and tpm, computed by summing over all transcripts for each gene
* (and `abundance.tsv` is renamed to `transcript_abundance.tsv`)

To get the `gene_abundance.tsv` we need a new input `gtf_gz`, the Ensembl GTF file for the host species that will tell it how to map the transcript IDs in `transcript_abundance.tsv` onto gene IDs for the roll-up. The input is optional and if absent then the `gene_abundance.tsv` output is omitted too.

Note: docker image update needed to install & upgrade some dependencies.


ovalenzuela19 (approved):

ovalenzuela19:
Thanks for this! Where can I find the GTF files per host genome?

PR #178, opened 2023-01-31
Auto download sra

rzlim08:
Auto-downloads file from SRA given an accession


rzlim08:
Cancelling since we've decided not to do this right away

PR #185, opened 2023-02-13
[SC-228638] Cleanup unused code

robertaboukhalil:
For exposing `total_reads`, `rpm`, and `dpm` in the AMR output file `combined_amr_results.csv`, I ended up going with a solution on the web app side, so I'm removing the code that was added to originally support this on the pipeline side.


rzlim08 (approved):
Thanks Robert!

PR #192, opened 2023-03-20
add unmapped read count

morsecodist:
The web application expects a count from unmapped reads but we weren't outputting one.


rzlim08 (approved):
LGTM

PR #193, opened 2023-03-20
[fix-ont-pz] Fix ONT pipeline viz bug

ovalenzuela19:


ovalenzuela19:
This change isn't necessary to fix the ONT PipelineViz so I'm closing this PR for now. I don't want to produce an unnecessary output if we don't have to

PR #195, opened 2023-03-22
Fix idseq-dag tests and revamp makefile

rzlim08:
* Not sure if people will consider this completely pointless, but I fixed the idseq-dag tests 
* IMO since we're planning on modifying/removing idseq-dag in the future it's good to fix these tests to ensure we don't accidentally modify or break anything downstream
* Also revamped the makefile for this repo. It looks like we've doubled down on `make` for the `czid-web` repo, so I thought it would be good to have an equivalent for this repo


robertaboukhalil (approved):
Looks good! Just a minor suggestion.

PR #199, opened 2023-03-28
Fix long reads errors

rzlim08:
* removes `expr` and uses bash arithmetic
* catches error in SeqIO reading fastq file
* quotes argument to `raise_error`


robertaboukhalil (approved):
Looks good!

PR #205, opened 2023-04-03
Output non host reads and non host contigs for AMR

ovalenzuela19:
Outputs the non host reads and non host contigs directly from the AMR workflow. After this change gets merged in we'll be outputting the non host reads and non host contigs files twice - one by itself and the other in the `outputs.zip` file.

I'll ask comp bio about what we should do about this - ideally we don't want to output the same file twice.
UPDATE: our cross functional partners said that it's ok to remove the non host contigs and reads from the outputs.zip

Their suggestion was to:
* Keep all the individual downloads
* Change “Download All with Intermediate Files (.zip)” to “Download Intermediate Files (.zip)” and only include the intermediate files in the download.


rzlim08 (approved):
Thanks!

ovalenzuela19:
@rzlim08 how do you concatenate the paired end non host reads files into just one fasta file?

rzlim08:
> @rzlim08 how do you concatenate the paired end non host reads files into just one fasta file? 

Ah, I didn't realize we were doing this for this PR. I have a relatively simple plan, will update this afternoon

workflows/amr/run.wdl
```diff
@@ -525,12 +521,10 @@ task ZipOutputs {
...
 
-        counter=1
-        for fastx in ~{sep= ' ' nonHostReads}; do 
-            cp $fastx non_host_reads_R$counter."${fastx#*.}"
-            ((counter++))
-        done
-        cp ~{sep=' ' outputFiles} .
+        # copy contigs and interleave non_host_reads
+        cp ~{contigs_in} contigs.fasta
+        seqfu ilv -1 ~{sep=" -2 " nonHostReads} | gzip > non_host_reads.fasta.gz
```
rzlim08, line 526: do I want to gzip this file?

rzlim08:
@ovalenzuela19 This should be ready now

workflows/amr/run.wdl
ovalenzuela19: I think the unzipped version is what the frontend expects, but maybe it's a good question for product/compbio. I think the gzipped version would be a good space saving solution

PR #206, opened 2023-04-06
[SC-230191] Filter spurious hits reads to contigs

rzlim08:
* Filters reads to contigs if the percentage clipped is >20%
* I'm hoping that writing this in python will be performant enough. We should test on a relatively large sample. 
* The challenging part of this was to stop python `csv` from mangling the quality scores on output. 

@katrinakalantar 
* This differs from your sample code as it only reads the head and tail of the `cigar` string. 
* The length is calculated using the length of the sequence, and adding the length of hard clipping as needed


lib/bin/filter_clipped_alignments.py
```diff
@@ -0,0 +1,66 @@
...
+def main(reads_to_contig_sam, output_file, max_percent):
+    with open(reads_to_contig_sam, "r") as csv_file, open(
+        output_file, mode="w", newline=""
+    ) as output_file:
+        csv_reader = csv.reader(csv_file, delimiter="\t")
+        output = csv.writer(
+            output_file,
+            delimiter="\t",
+            lineterminator="\n",
+            quotechar="\t",
```
rzlim08, line 37: this and `QUOTE_NONE` were necessary for the csv module to stop mangling the quality string in case of quotation marks `"`

lib/bin/filter_clipped_alignments.py
```diff
@@ -50,6 +65,14 @@ def main(reads_to_contig_sam, output_file, max_percent):
                 percent_clipped = ((front_clipping + end_clipping) / alength) * 100
                 if percent_clipped < max_percent:
                     output.writerow(row)
```
robertaboukhalil, line 67: Out of scope for this PR, but it might be useful to write this script using the [pysam library](https://pysam.readthedocs.io/en/latest/api.html) so you don't have to deal with column numbers, cigar parsing, TSV parsing, etc. It also lets you open a sam/bam file for reading and another one for writing!

rzlim08: Thanks @robertaboukhalil I realized this like 80% through haha

katrinakalantar (commented):
This looks good to me! I see the 20% threshold in the script call now.

PR #208, opened 2023-04-07
[SC-230372] Optionally output the primer for the consensus genome pipeline

rzlim08:
* Pipeline part for optionally outputting the primer
* Webapp part needs to figure out if the user submitted the primer then turn the flag on


phoenixAja (approved):
much cleaner solution than what i envisioned! thank you

cc @irene-han once we transition back to WGS we'll need to update the upload endpoint accordingly

PR #213, opened 2023-04-12
TSV to SAM with pysam

robertaboukhalil:
Rewriting my previous logic for converting `comprehensive_AMR_metrics.tsv` to SAM so that it uses `pysam` instead of awk 🙂


rzlim08 (approved):
Looks good, thanks!

PR #216, opened 2023-04-18
[Bug] fix count expansion for single file short-read-mngs

rzlim08:
* Fix count expansion for single file workflows


ovalenzuela19 (approved):

ovalenzuela19:
Thanks so much for fixing these issues!

PR #217, opened 2023-04-19
modern host filtering: use sorted outputs of Bowtie2 and HISAT2

mlin:
For repeatability: back/forward-porting #171 to the modernized host_filter.wdl

We already had a *background* task to sort the Bowtie2 outputs (to generate an archival BAM), so this just waits for & uses those sorted reads in the following steps. Increased the thread count for sorting since we now block on it.

Add similar sorting step to HISAT2 (but we don't output the BAM in this case).


mlin:
@katrinakalantar @rzlim08 @ovalenzuela19 deferring to you on whether this is release-blocking

mlin:
@katrinakalantar @rzlim08 @ovalenzuela19 I had to change one of the taxid's used in the non-host alignment unit tests that was no longer hit as a result of this change.

I was concerned about this because that taxid has been hardcoded in the test cases for years, with both the original and modernized pipeline. But, I carefully verified that the modern host filtering pipeline is generating exactly the same reads with and without this patch, just in a different order; so as best I can tell, it seems to be a true case of the read order affecting the results (which of course is the reason for this patch in the first place). 

I'm still surprised, though, so due diligence may call for a little extra testing in staging...

rzlim08 (approved):
Looks good!

rzlim08:
Hi Mike, this looks good to merge, thanks for taking care of this. 

Looks like the squash of the previous branch made this a bit complicated, but I ran

```git rebase --onto main mlin/modernize-host-filter```

And copied it into [here](https://github.com/chanzuckerberg/czid-workflows/pull/233) if it makes it easier to merge!

PR #233, opened 2023-04-28
modern host filtering: use sorted outputs of Bowtie2 and HISAT2

rzlim08:
* Rebase of https://github.com/chanzuckerberg/czid-workflows/pull/217


mlin (approved):
Thanks @rzlim08 

Copying this to preserve in the history:

> I had to change one of the taxid's used in the non-host alignment unit tests that was no longer hit as a result of this change.

> I was concerned about this because that taxid has been hardcoded in the test cases for years, with both the original and modernized pipeline. But, I carefully verified that the modern host filtering pipeline is generating exactly the same reads with and without this patch, just in a different order; so as best I can tell, it seems to be a true case of the read order affecting the results (which of course is the reason for this patch in the first place).

> I'm still surprised, though, so due diligence may call for a little extra testing in staging...

mlin:
Oh, retrying / looking into the test failure.

rzlim08:
> Oh, retrying / looking into the test failure.

Thanks! I think I just left in the `wait` line in bowtie2_filter

PR #240, opened 2023-05-10
Update ERCC count output name

ovalenzuela19:
Renames the ERCC count output name. 
 `ercc_bowtie2_filter_count` -> `bowtie2_ercc_filtered_out_count`

This old count output name `ercc_bowtie2_filter_count` strays from the pattern we've been using to name the SFN output counts `bowtie2_ercc_filtered_out_count`


rzlim08 (approved):
Thanks!

PR #243, opened 2023-05-11
separate bed and refseq download params

phoenixAja:
bed files are optional on upload for WGS, separate the logic to download refseq and bed file together


phoenixAja:
it seems like the integration test is failing on downloading the hmmer and fasta3 packages and not anything that i did here?

j-x-han (approved):

rzlim08:
Nice! Thanks Phoenix!

PR #244, opened 2023-05-15
[CZID-7837] Exit early if the assembly failed

robertaboukhalil:
This PR fixes the bug encountered in CZID-7837.

When the SPADES assembly fails, we create a `contigs.fasta` file containing `;ASSEMBLY FAILED`, which is not a valid FASTA. This leads the `TsvToSam` step to fail when it tries to create an FAI index for `contigs.fasta`. Instead I'm checking for this special case and exiting the TsvToSam step early.

### Tests

* Tested on staging that the sample previously failing is no longer failing
* Also tested that samples that previously did not fail are still not failing

### DIFF
--- a/workflows/amr/run.wdl
+++ b/workflows/amr/run.wdl
@@ -636,19 +636,20 @@ task tsvToSam {
         python3 <<CODE
         import pandas as pd
         import pysam
+        import sys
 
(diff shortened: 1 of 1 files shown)


robertaboukhalil:
cc @rzlim08: just fyi for when you're back

ovalenzuela19 (approved):
Thanks so much Robert!!

PR #248, opened 2023-05-18
Fix bug in long-read-mngs filter clipped alignments

morsecodist:
We were getting the following error on a user's sample from this script:

```
 _csv.Error: field larger than field limit (131072)
 ```

Based on [this](https://stackoverflow.com/questions/15063936/csv-error-field-larger-than-field-limit-131072) answer it looks like there is either a delimiter problem or the field is actually just too big. I checked the file in question and it does have a contig that long. That makes sense, contigs can be quite long. Also confirmed it worked with the size upgrade.

### DIFF
--- a/lib/bin/filter_clipped_alignments.py
+++ b/lib/bin/filter_clipped_alignments.py
@@ -1,6 +1,18 @@
 import csv
 import re
 import argparse
+import sys
+
+# the SAM file input has the full sequence which can be very long
+#   increase the csv field size limit to ensure they will fit
+_field_size_limit = sys.maxsize
+while True:
+    try:
+        csv.field_size_limit(_field_size_limit)
+        break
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/bin/filter_clipped_alignments.py [hunk 1]
```diff
@@ -1,6 +1,18 @@
...
+import sys
+
+# the SAM file input has the full sequence which can be very long
+#   increase the csv field size limit to ensure they will fit
+_field_size_limit = sys.maxsize
+while True:
+    try:
+        csv.field_size_limit(_field_size_limit)
+        break
+    # if the size limit is two high make it one bit smaller and try again
```
irene-han, line 13: Nit: "too high"

irene-han (approved):

PR #257, opened 2023-05-30
Fix fastp error if no reads after ercc filtering

rzlim08:
* Fastp crashes if given an empty input file
* handles empty output from ercc bowtie step


robertaboukhalil (approved):
Nice!

PR #261, opened 2023-06-02
Make sure kallisto is run single-threaded (with a fixed seed)

lvreynoso:
Specify `--threads=1 --seed=42` when running `kallisto`.

See CZID-7785.

Successful local run:
Screenshot 2023-06-05 at 12 09 48


rzlim08 (approved):
Nice! Looks good!

PR #263, opened 2023-06-06
Properly escape files with spaces

rzlim08:
* we removed the block on spaces and parentheses in input files, so need to escape things properly


phoenixAja (approved):
cool! A few things:
*  Did you test this out by chance? 
* Do you think we can release this before noon today? 
* Do we also have to worry about escaping ()?

phoenixAja:
I'm also working on a fix for this on the FE side by surfacing a tooltip for any other characters

rzlim08:
> cool! A few things:


> * Did you test this out by chance?
* yes locally I did

> * Do you think we can release this before noon today?
* Yep!
> * Do we also have to worry about escaping ()?
* I also tested escaping ()

PR #270, opened 2023-06-22
Sort input to insert_size_metrics

rzlim08:
* sorts inputs to `picard CollectInsertSizeMetrics` by coordinate 
* error due to running out of space. but will fix later.

### DIFF
--- a/workflows/short-read-mngs/host_filter.wdl
+++ b/workflows/short-read-mngs/host_filter.wdl
@@ -860,7 +860,7 @@ task collect_insert_size_metrics {
 
   command <<<
     samtools sort -o "bowtie2_coordinate_sorted.bam" -@ 8 -T /tmp "~{bam}"
-    picard CollectInsertSizeMetrics 'I=bowtie2_coordinate_sorted.bam' O=picard_insert_metrics.txt H=insert_size_histogram.pdf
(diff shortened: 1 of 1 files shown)


workflows/short-read-mngs/host_filter.wdl
```diff
@@ -859,7 +859,8 @@ task collect_insert_size_metrics {
   }
 
   command <<<
-    picard CollectInsertSizeMetrics 'I=~{bam}' O=picard_insert_metrics.txt H=insert_size_histogram.pdf
+    samtools sort -o "bowtie2_coordinate_sorted.bam" -@ 8 -T /tmp "~{bam}"
+    picard CollectInsertSizeMetrics 'I=bowtie2_coordinate_sorted.bam' O=picard_insert_metrics.txt H=insert_size_histogram.pdf
```
vincent-czi, line 863: Nit: Having the single ticks wrapping `'I=bowtie2_coordinate_sorted.bam'` seems unnecessary at this point and a little confusing since the other args don't have matching ones. I assume it was there before because of the `~` interpolation? If you agree and remove it here, we should also remove it from the documentation below too.

vincent-czi (approved):

workflows/short-read-mngs/host_filter.wdl
rzlim08: Thanks Vincent, that's a good call. I've removed this from both places.

PR #274, opened 2023-07-07
Add high level drug class to AMR output

lvreynoso:
Adds a new input to the AMR workflow and the RunResultsPerSample task within it for a json file with ontology information. Queries the json file for high level drug classes for a given gene name and adds them in a new column in `primary_AMR_report.tsv`.

Results from a successful run:
Screenshot 2023-07-07 at 14 23 10

After I made this screenshot I modified the `.join` call so that high level drug classes would have a space after each semicolon, like in the drug_class column.

### Notes
A previous commit added a call to `seqkit` but used flags found in a later version than the one installed. I updated the `seqkit` version in the amr dockerfile so the expected behavior would happen.

### DIFF
--- a/scripts/docker-build.sh
+++ b/scripts/docker-build.sh
@@ -2,4 +2,4 @@
 
 set -euxo pipefail
 
-docker buildx build --build-context lib=lib "$@"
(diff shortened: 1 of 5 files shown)


rzlim08 (approved):
This looks great!

PR #275, opened 2023-07-10
Handle case where gene entry in ontology has no HLDC

lvreynoso:
CZID-8052

Handles the case where a gene has an entry in the ontology file, but the gene's entry is missing the key `'highLevelDrugClasses'`.

Modifies test data to take this case (as well as a missing gene entry in the ontology file) into account.

No longer drops rows without a `high_level_drug_class` entry.


ovalenzuela19 (approved):
LGTM! Tested this case on web app

rzlim08 (approved):

PR #278, opened 2023-07-17
Use drug class instead of gene name to query for high level class

lvreynoso:
Obtains the names of high-level drug classes for each gene by using the gene's drug class to query the ontology file.

Sample output:
Screenshot 2023-07-17 at 17 07 33

### DIFF
--- a/workflows/amr/run.wdl
+++ b/workflows/amr/run.wdl
@@ -283,11 +283,14 @@ task RunResultsPerSample {
             return([i for i in set_list if i == i])
 
         ontology = json.load(open("~{card_ontology}"))
-        def get_high_level_classes(drug_classes):
-            high_level_classes = []
+        def get_high_level_classes(drug_classes_union):
+            drug_classes = set()
+            for drug_class_string in drug_classes_union:
+                drug_classes |= set(drug_class_string.split('; '))
+            high_level_classes = set()
             for drug_class in drug_classes:
                 if drug_class in ontology:
-                    high_level_classes.extend(ontology[drug_class]['highLevelDrugClasses'])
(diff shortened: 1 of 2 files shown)


workflows/amr/run.wdl
```diff
@@ -283,12 +283,15 @@ task RunResultsPerSample {
             return([i for i in set_list if i == i])
 
         ontology = json.load(open("~{card_ontology}"))
-        def get_high_level_classes(gene_name):
-            if gene_name not in ontology:
-                return []
-            if 'highLevelDrugClasses' not in ontology[gene_name]:
-                return []
-            return ontology[gene_name]['highLevelDrugClasses']
+        def get_high_level_classes(drug_classes_union):
```
ninabernick, line 286: what's the form of `drug_classes_union`? A set of strings like "rifamycin antibiotic; peptide antibiotic"? Wondering why it would have length > 1 ever

lvreynoso: It's type is `list[str]` but there is usually only one element in the list, a string containing drug classes separated by semicolons, e.g.:

Screenshot 2023-07-18 at 09 15 36

The variable is created by this line of code:
```python
dc = remove_na(set(sub_df['Drug Class_contig_amr']).union(set(sub_df['Drug Class_kma_amr'])))
```
This can result in two elements in the list if the two columns differ.

lvreynoso: I'm hoping that soon I can move the python code out to a separate file and upgrade the container's python version to >= 3.9 so type hinting is a lot easier to add.

lvreynoso: In the meantime that's probably a misleading variable name, I'll change it to something better

ninabernick (approved):

PR #279, opened 2023-07-18
Use regex to ensure only final _\d sequence in contig names is dropped

lvreynoso:
Fixes a string replacement function for contig names so that only any extraneous `_1`, `_2,` etc suffixes are dropped. Previously, everything after the final underscore would be dropped.

Tested on an EC2 instance.


ninabernick (approved):
Great catch!

PR #281, opened 2023-07-19
Measure AMR contig % coverage by using coordinate ranges

lvreynoso:
Changes the calculation of contig gene coverage in the `MakeGeneCoverage` task by aggregating contig coordinates in a Python set. This allows us to calculate the percentage of the reference genome covered by the gene's contigs in sample regardless of the orientation of the contigs or the presence of a paired set of contigs.

### Notes

The output gene coverage table before, after processing an unpaired reverse contig for a gene (Row 1):

Screenshot 2023-07-19 at 16 00 25

And after the changes in the PR:

Screenshot 2023-07-19 at 15 58 22

I tested this branch on an EC2 instance as well.

### DIFF
--- a/workflows/amr/run.wdl
+++ b/workflows/amr/run.wdl
@@ -621,21 +621,25 @@ task MakeGeneCoverage {
     agg_res = df.groupby(["ARO", "Model_ID"]).agg(lambda x: list(x))
 
     gene_coverage = []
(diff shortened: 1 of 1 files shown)


rzlim08 (approved):
Thanks this looks great! I guess I didn't know that the contig could be reversed

PR #285, opened 2023-08-09
Mapped+Unmapped fix for short-read-mngs

vincent-czi:
Fix for CZID-8231.

# Summary
We had an issue where paired end files were not coming out with the same number of `1` and `2` reads when downloading non-host files. This fixes that so Non-Host downloads now have matched numbers of reads, and more generally fixes our Non-Host downloads so they have mapped+unmapped reads in them, as users expect.

## Details: Cause
The underlying cause behind this issue was that we were filtering non-host reads down to only those reads that could be _mapped_ to a specific accession (this happened in `GenerateAnnotatedFasta`). The file being generated with mapped-only reads was `refined_taxid_annot.fasta`. However, that wasn't a bug, per se: we were doing this on purpose because a later task (`GenerateTaxidLocator`) needs to receive _only_ mapped reads. If that task gets unmapped reads, it causes weird behavior in the sample report where the unmapped reads get lumped in where they shouldn't really be.

The problem is that users expect their Non-Host Reads downloads to be _all_ reads that were non-host, whether those reads were mapped or unmapped. This should apply to all non-host downloads, both the fasta and fastq(s), regardless of the paired end aspect. This meant that we had a logical inconsistency where `refined_taxid_annot.fasta` was expected to be two, mutually exclusive things: in the workflow, it was expected to be mapped-only, while from the webapp's POV, it was supposed to be mapped+unmapped.

As for how this connects to the paired reads aspect in the original ticket, that's pretty much just how we first noticed this issue. Here's how it came up in a paired reads flow:
- The original, raw files from the user start out with every read being paired.
- A read being unmapped is effectively random, and it's not locked to whether or not its pair-buddy is mapped or not. You can have a paired read where `/1` is unmapped while `/2` is mapped, or vice-versa.
  - ^^^ The fact that pairs are not locked together is surprising! But I confirmed with Katrina that this is the known and expected behavior for our app.
- Because the unmapped trait is effectively random, in most cases there isn't a perfectly equal number of unmapped in the `/1` and `/2` categories.
- Thus the mapped-only `refined_taxid_annot.fasta` would (most of the time) not have a matched number of `/1` and `/2` reads. This then percolated on to the other Non-Host download files of `nonhost_R1.fastq` and `_R2`, so those did not match up as well.

But ultimately, this wasn't really a paired reads issue, it was a mapped-only vs mapped+unmapped issue. It just so happens that we noticed that things weren't paired up anymore, and the reason behind that was unmapped reads weren't appearing. 

## Details: Fix
The fix was fairly simple: just start creating two fastas, one mapped-only and another for mapped+unmapped.

To avoid needing to change anything in the webapp code, The **`refined_taxid_annot.fasta` now is the mapped+unmapped** file. By changing the contents of that file to be mapped+unmapped but leaving the name alone, the webapp will now fetch the mapped+unmapped reads when the user selects a Non-Host reads download, as they expect. (This also percolates to the `nonhost_R1.fastq` and `_R2` files since those get built using that fasta.)

To keep the workflows behaving as expected, there is now a **new file of `refined_taxid_annot_mapped_only.fasta` that is mapped-only** reads. This then gets passed on to the `GenerateTaxidLocator` task that assumes it only receives mapped reads, so everything continues to work as desired. It will be available in S3 if we need it after a workflow run, but it shouldn't appear to the user unless they specifically go looking for it in the pipeline results folder.

As for the creation of the mapped+unmapped file, that's pretty simple. It just extends the previous creation process for mapped-only to add on the unmapped reads as well in a new file. After mapped-only creation is done, it copies that file as a starting place, then appends on all the unmapped (AKA, unidentified) reads while conforming their headers to match the taxid structure we expect elsewhere from reads in that file.

## Other
- There's been a long running linter error that is totally unrelated to this work. We've had it failing on a bunch of GHA runs for the last couple weeks. This was in `lib/idseq-dag/idseq_dag/util/m8.py`, fix was very simple.
- In the `NonhostFastq` task, we had an additional call of `run_with_tax_ids` that was specifically targeting betacoronaviruses, but as far as I could tell, we ran this then immediately dropped the outputs, it did nothing of impact at all. Best guess is that it was added in early on with COVID for a specific bit of research work, but has not been used for quite awhile now. I removed that code and also refactored the functions there to simplify them. Supporting a taxid[s]-specific run (eg, betacoronaviruses) added extra complexity to the underlying functions. With that gone, there was no reason to keep that complexity around, so I refactored it out.
  - I did a run of the `NonhostFastq` task before and after the commit that refactored the `run_with_tax_ids` stuff. I confirmed that the output was unchanged with the refactor, just like expected.
  - There is a slim possibility that I'm somehow missing a way that these outputs get exposed and there is someone at CZ BioHub still using it. This seems very unlikely, but if it happens, we can just revert the code in `lib/idseq-dag/idseq_dag/steps/nonhost_fastq.py` to what it was before commit 8c322a9.

### DIFF
--- a/lib/idseq-dag/idseq_dag/steps/generate_taxid_fasta.py
+++ b/lib/idseq-dag/idseq_dag/steps/generate_taxid_fasta.py
@@ -1,4 +1,5 @@
 from typing import List
+from shutil import copy as file_copy
 from idseq_dag.util.parsing import HitSummaryMergedReader
 from idseq_dag.engine.pipeline_step import PipelineStep
 import idseq_dag.util.lineage as lineage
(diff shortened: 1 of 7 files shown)


lib/idseq-dag/idseq_dag/steps/generate_taxid_fasta.py
```diff
@@ -56,18 +57,64 @@ def generate_taxid_fasta(
...
     """Generate taxid FASTA from hit summaries. Intermediate conversion step
     that includes handling of non-specific hits with artificial tax_ids.
     """
 
     def run(self):
         input_fa_name = self.input_files_local[0][0]
-        if len(self.input_files_local) > 1:
+        # We determine if intended for use in `short-read-mngs/postprocess.wdl`
+        # or `short-read-mngs/experimental.wdl` by shape of input files list.
+        is_structured_as_postprocess_call = (len(self.input_files_local) > 1)
```
rzlim08, line 113: nit: would it be better to test with the length of the output files array?

rzlim08 (approved):
Looks good just a small nit!

lib/idseq-dag/idseq_dag/steps/generate_taxid_fasta.py
vincent-czi: I suppose we could use that now, I was just transforming the previous way of checking to try to remain as consistent with the old code as possible. I agree that length of outputs is probably a cleaner check, but I'm just wary to make any material changes just in case I missed something else.

PR #287, opened 2023-08-16
Re-add duplicates to AMR pipeline

lvreynoso:
[CZID-7985](https://czi-tech.atlassian.net/browse/CZID-7985)

[CZID-7985]: https://czi-tech.atlassian.net/browse/CZID-7985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

This PR re-adds duplicate reads that have been removed via the host filtering stage. This affects workflows that use `raw_reads_0` and `raw_reads_1` as sample input, i.e. in CZID samples that are uploaded directly for AMR analysis. This does not affect samples that were originally uploaded to other workflows (mNGS or ONT) since the `non_host_reads` and `contigs` inputs assume that they have duplicates already added back in. The most dramatic differences will be seen in FLASH samples.

The code here accomplishes this by taking in reads remaining after HISAT2 filtering in the host stage, as well as the duplicate cluster ids and sizes analysis from the dedup stage, and comparing them to the reads left after subsampling. Reads that remain after host filtering and are identical to one in the subsampled set of reads are interleaved with the subsampled reads before being fed to the rest of the workflow.

If the redup step sees that there are no cluster sizes >1 in the dedup output, it simply outputs the subsampled reads.

Example of a FLASH sample run on the current workflow:
Screenshot 2023-09-11 at 10 18 10

Same sample run on the version in this PR:
Screenshot 2023-09-11 at 10 18 20


rzlim08 (approved):
This looks good but is there a case where we'll need to re-add the duplicates from the short-read-mngs side? I.e. when we're kicking off from a sample that has already been host filtered?

lvreynoso:
> This looks good but is there a case where we'll need to re-add the duplicates from the short-read-mngs side? I.e. when we're kicking off from a sample that has already been host filtered?

I looked into that last week and thought that wouldn't be an issue, but I looked again after reading your comment and it appears I made a mistake. That case does exist for short-read mNGS samples. Currently, the webapp uses the output of the short-read mNGS subsampling step as input to the AMR workflow, along with the contigs fasta output from the short-read mNGS postprocessing step. The task to generate contigs does use duplicates as part of its input, but the subsampled file is stripped of all duplicates. The context needed to reinsert duplicate reads is not present.

This PR would only affect samples uploaded directly to the AMR workflow, so samples already run on the short-read mNGS pipeline would not see any benefit.

However, there is a simple solution: update the webapp to use the nonhost fastq files generated by the `NonhostFastq` task in the `experimental` workflow instead of the subsampled task output. I will ask if this is something we can do, because that would make things much easier (and require no code changes to this PR).

If that is not the case and we wanted to avoid rerunning host-filtering, we would have to add four more parameters to the workflow for non-host reads before deduplication and subsampling, cluster ids and cluster sizes, and a flag to distinguish mNGS samples from ONT samples, logic to deal with those samples, and logic on the webapp side to include those files when running short-read mNGS samples on this workflow version and up.

PR #289, opened 2023-08-25
Fix Benchmarking 'NoneType' object is not subscriptable error

valenzuelaomar:
When `cur` is `None`, the function `load_contig_lengths` will fail when trying to index on the None type: `cur[0], cur[1]`. This fixes that issue by checking if the contigs assembly failed before computing contigs length


rzlim08 (approved):
Thanks!

PR #293, opened 2023-09-01
Convert Kallisto gene count to transcript count

vincent-czi:
# Description

CZID-8315

This PR stops the production of the `reads_per_gene.kallisto.tsv` file to instead focus on making the `reads_per_transcript.kallisto.tsv` file more accessible to our users. My [recent PR in the webapp](https://github.com/chanzuckerberg/czid-web-private/pull/3829) changes the bulk download of host count info to target downloading reads_per_transcript instead of _gene. For comp bio reasons, we no longer want to do the rollup to gene counts ourselves, so we're stopping creation of those files. Additionally, to better enable users to do analysis on the transcript files, with this PR we now also create a mapping TSV that shows how each transcript id maps to its corresponding gene id.

Overall intent is for users to more easily be able to download the transcript TSVs (they were in the results folder before, but now they're also available via Bulk Download [although that work already happened in above linked webapp PR]), have access to the transcript_id <--> gene_id mapping (just via results folder is fine), and stop producing the gene count TSV in general.

## Notes

- @rzlim08 @morsecodist With this PR, the `gtf_gz` input no longer does anything anywhere. To avoid any worries about payloads from the webapp containing that param and breaking the run because the param is not accepted, I've removed it from the task, but left it in at the WDL `workflow` level as an optional. Can you look at this part in specific and make sure it seems sane?
- Similarly to the above, we were using a Python library called `gtfparse` to consume the GTF file in the Kallisto task. But since we no longer do the roll-up, we no longer use the GTF anymore, so we can drop the library. If anyone thinks it's a bad idea to drop this library being installed in the Dockerfile, please let me know! But I can't see a reason to keep it around any more.
  - Sidebar: Confusingly, `gtfparse` also gets installed in the `legacy-host-filter` Dockerfile, but that WDL also has no usage of gtfparse? Or anything GTF or Kallisto? Dunno why it ever appeared over there, but I figure I'll leave that one alone.
  - I guess it's possible that `gtfparse` is a dep of one of our deps, and we do specify its version in the Dockerfile, so it's possible unpinning the gtfparse version could cause problems? I'm just trying to be a devil's advocate here for how removing it could go wrong.


workflows/short-read-mngs/host_filter.wdl
```diff
@@ -175,7 +176,7 @@ workflow czid_host_filter {
 
     File kallisto_transcript_abundance_tsv = kallisto.transcript_abundance_tsv
     File kallisto_ERCC_counts_tsv = kallisto.ERCC_counts_tsv
-    File? kallisto_gene_abundance_tsv = kallisto.gene_abundance_tsv
+    File kallisto_transcript_gene_mapping_tsv = kallisto.transcript_gene_mapping_tsv
```
rzlim08, line 179: Is this an optional file? Will this be produced on every pipeline run?

rzlim08 (approved):
This looks reasonable, though I don't have much knowledge of the transcript_to_gene mapping or reads_per_gene mapping. 

The gtf_gz file seems like it's ok like that there will be a small warning if running `miniwdl check` but I don't think it should have a functional impact. 

We might need to add a ? in the new `transcript_to_gene` file if it's optional

workflows/short-read-mngs/host_filter.wdl
vincent-czi: I don't think it's an optional file, because the `kallisto.transcript_abundance_tsv` is produced on every run as far as I can tell, and `kallisto.transcript_gene_mapping_tsv` is a direct child of that file that we want to produce for every transcript TSV.

rzlim08: ok make sense, thanks!

PR #297, opened 2023-10-13
[CZID-8579] Fix typo

robertaboukhalil:
Fix a small typo that causes the warning log to error out. [Docs](https://docs.python.org/3/library/logging.html#logging.LogRecord)

```
"TypeError: _log() got an unexpected keyword argument 'exec_info'"
```


rzlim08 (approved):
thanks for catching this!

PR #298, opened 2023-10-16
[CZID-8703] unit test lineages csv generation

ninabernick:
Jira ticket: [CZID-8703](https://czi-tech.atlassian.net/browse/CZID-8703)

Implements a series of tests for creating a versioned CSV file in `version_taxon_lineages`. Takes in a previous CSV file, which contains the old taxon lineages table, and a current lineages file generated using `ncbitax2lin.py`. 

Also incorporates changes from @morsecodist that fix a major bug with the lineage generation: only one existing row for each entry was being copied over to the new lineages table, and other deprecated rows were being dropped.

For each test case (1 case = 1 taxon), we test how many rows are in the output CSV and what the `version_start` and `version_end` fields are. Also tests total number of rows in output.

Test cases:
- unchanged taxon
- unchanged taxon with multiple existing rows
- taxon with multiple existing rows that is changed a second time
- taxon missing genus is reassigned to new family
- taxon has simple genus change
- taxon is deprecated
- taxon with multiple existing rows is deprecated
- taxon changed from phage to non-phage
- new taxon is added

[CZID-8703]: https://czi-tech.atlassian.net/browse/CZID-8703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


workflows/index-generation/generate_lineage_csvs.py
```diff
@@ -184,24 +206,26 @@ def version_taxon_lineages(
...
-                previous_row = previous_lineages.get(row["taxid"])
-
-                if previous_row:
-                    # If we have a previous row, remove it so we can later
-                    #   iterate through only the rows from previous_lineages
-                    #   that aren't also in lineages
-                    del previous_lineages[row["taxid"]]
+                previous_row = previous_lineages.pop(
+                    (row["taxid"], previous_lineages_version), None
+                )
```
valenzuelaomar, line 228: To clarify my understanding, the bug we had previously was caused by us only keeping track of the `previous_row` based on `tax_id` and deleting the previous row later on when we compare against the new lineage. This approach didn't allow more than one previous taxon row from being recorded, thus dropping deprecated rows 

This bug fix adds the `version_end` to the `previous_lineages` dictionary in addition to the `tax_id` so that we can keep track of the deprecated rows.

Is my understanding of the issue correct?

workflows/index-generation/generate_lineage_csvs.py
```diff
@@ -184,24 +206,26 @@ def version_taxon_lineages(
...
     previous_lineages = {}
+    previous_lineages_version = None
     if previous_lineages_filename:
         with gzip.open(previous_lineages_filename, "rt") as f:
             for row in csv.DictReader(f):
-                previous_lineages[row["taxid"]] = row
+                previous_lineages[(row["taxid"], row["version_end"])] = row
+                if (
+                    not previous_lineages_version
+                    or previous_lineages_version < row["version_end"]
```
valenzuelaomar, line 216: pretty cool that the string comparison works well with these operators, didn't know python could do that!

workflows/index-generation/generate_lineage_csvs.py
ninabernick: Yes, that's exactly correct.

PR #299, opened 2023-10-20
[CZID-8703] [CZID-8778] log taxa and number of lines in index generation output

ninabernick:
Jira ticket: CZID-8703

This adds logging to report the number of lines corresponding to different changes in taxonomy:
- unchanged lineages (should be copied over and have `version_end` updated
- updated lineages (should have new row added in new table)
- new taxa
- deprecated rows (includes both outdated lineages for existing taxa and deprecated taxa, deprecated means `version_end` is not updated)
- total rows written to new table

It also logs the following groups to CSV files:
- new taxa
- deprecated taxa
- updated taxa (indicates level at which it was changed)

Count variables are used because we don't load the whole output into memory at once.

Asserts that:
- total # of rows = # existing rows + # updated taxa + # new taxa
-  total # existing rows = # deprecated rows + # unchanged rows


workflows/index-generation/generate_lineage_csvs.py
```diff
@@ -141,6 +144,30 @@ def _equals(previous_row: Dict[str, str], row: Dict[str, str]):
     return True
 
 
+def _find_highest_taxonomy_changed(previous_row: Dict[str, str], row: Dict[str, str]):
+    """
+    Finds the highest rank taxonomy level at which a taxon was reclassified. e.g. if species A was reassigned
+    to a different genus and family, it would return the new family name instead of the genus name.
+    """
+    fieldnames_to_search = [
```
rzlim08, line 152: Could you use `_fieldnames` here?

workflows/index-generation/generate_lineage_csvs.py
```diff
@@ -242,17 +310,84 @@ def version_taxon_lineages(
...
                     row["updated_at"] = str(datetime.now())
-                    writer.writerow(row)
+                    versioned_csv_writer.writerow(row)
+                    num_total_new_rows += 1
 
                     if previous_row:
-                        writer.writerow(previous_row)
+                        # this is an updated lineage
+                        (
+                            taxonomy_level,
```
rzlim08, line 319: This value may not be a `taxonomy_level` (e.g. `is_phage`)

rzlim08 (approved):
Looks good! Just a couple of nits

workflows/index-generation/generate_lineage_csvs.py
ninabernick: The reason I didn't was that I wanted to traverse them in a particular order (level name, level id, is phage, tax id, tax name) from highest to lowest rank. This is so that we find the highest level at which the taxonomy changed

PR #304, opened 2023-11-06
check if phage name contains 'phage'

ninabernick:
Checks if virus name contains "phage" when assigning `is_phage`. Captures case of unclassified bacterial phage without family.


rzlim08 (approved):
Nice, are we relatively sure that everything named "phage" is a phage? Like there's no e.g. taxa named "phage-like" or anything?

rzlim08:
> Nice, are we relatively sure that everything named "phage" is a phage? Like there's no e.g. taxa named "phage-like" or anything?

Actually would we mis-classify the clade `["aphagea"](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=2704112&lvl=3&p=has_linkout&p=blast_url&p=genome_blast&p=mapview&lin=f&keep=1&srchmode=2&unlock)`?

ninabernick:
> > Nice, are we relatively sure that everything named "phage" is a phage? Like there's no e.g. taxa named "phage-like" or anything?
> 
> Actually would we mis-classify the clade ["aphagea"](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=2704112&lvl=3&p=has_linkout&p=blast_url&p=genome_blast&p=mapview&lin=f&keep=1&srchmode=2&unlock)?

Great catch! Looks like the old query was checking for a whole word match to "phage", I'll fix that

PR #305, opened 2023-11-09
[CZID-8624] Enforce and automate changelog updates

morsecodist:
This enforces developers update a changelog and automates the versioning of the changelog. When developing developers can now update the changelog with unreleased notes like:

```
### short-read-mngs-unreleased
- my notes
```

If they don't, the PR check fails. At deploy time the correct version is automatically filled in instead of unreleased.

### DIFF
--- a/.github/workflows/release_workflows.yml
+++ b/.github/workflows/release_workflows.yml
@@ -9,7 +9,7 @@ on:
         description: major, minor, or patch
         default: patch
       release_notes:
-        description: A text message summarizing the changes in this release
+        description: A text message summarizing the changes in this release, only used for branch releases
 
 env:
   GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -28,19 +28,30 @@ else
     exit 1
 fi
 
+<<<<<<< Updated upstream
 # Fix until we can update this script to follow semantic versioning
 if [[ $WORKFLOW_NAME == amr ]]; then
     TAG="amr-v1.3.0"
 fi
+=======
+TAG_MSG=$(mktemp)
+>>>>>>> Stashed changes
 
 if [[ $( git branch --show-current) != "main" ]]; then 
     COMMIT=$(git rev-parse --short HEAD)
     TAG=$TAG"-$COMMIT"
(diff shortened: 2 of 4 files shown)


scripts/release.sh
```diff
@@ -32,15 +32,23 @@ fi
 if [[ $WORKFLOW_NAME == amr ]]; then
     TAG="amr-v1.3.0"
```
vincent-czi, line 33: Unrelated to this PR, but is amr just locked down 1.3.0 entirely at the moment? What's going on here?

.github/workflows/release_workflows.yml
```diff
@@ -9,7 +9,7 @@ on:
         description: major, minor, or patch
         default: patch
       release_notes:
-        description: A text message summarizing the changes in this release
+        description: A text message summarizing the changes in this release, only used for branch releases
```
vincent-czi, line 12: I was a bit confused by this phrasing until you explained it on the team meeting. Maybe expand this a bit to say:
"A text message summarizing the changes in this release. Unused for standard major/minor/patch releases, but used for branch releases."

I'm not sure that's great, but just something clarifying that it doesn't have an expected use in "normal" release flows.

vincent-czi (commented):
Is there a plan to convert over to **only** using the top-level `CHANGELOG.md` file? Because we also have `# Changelog` sections of the READMEs for both long-read and AMR workflows. It looks to me like this approach assumes that there is only a single changelog file for all of the workflows. If we're going to standardize on a single changelog file for all workflows in the upcoming backfill PR to get our changelogs up to date, no issue. But if that's not planned, uhhh, problems.

scripts/release.sh
lvreynoso: Hey Vincent, this ugly hack was me. This release script doesn't distinguish between a beta and a full release tag, so something like `amr-v1.3.0-beta.1-0xasdf` looks exactly the same as `amr-v1.3.0` to it, which is an issue when I want to release a beta version of the workflow so I can test it in sandbox or staging, because if I do a branch release of `1.3.0-blahblah` then when I merge the changes into main the script bumps up the patch/minor/version number and tags the full release as `1.3.1`.

I want to update the script but haven't gotten around to it.

vincent-czi: Got it, thanks!

PR #306, opened 2023-11-18
[CZID-8914] Mark empty SPAdes output as failed assembly

lvreynoso:
Fixes a bug where the workflow failed to handle cases in which samples ran through SPAdes assembly without error but did not have any contigs. The SPAdes output was not marked as ";ASSEMBLY FAILED", causing an error further along in the workflow.


scripts/release.sh
```diff
@@ -30,7 +30,7 @@ fi
 
 # Fix until we can update this script to follow semantic versioning
 if [[ $WORKFLOW_NAME == amr ]]; then
-    TAG="amr-v1.3.1"
+    TAG="amr-v1.3.2-beta.1"
```
rzlim08, line 33: To be honest, I think we can remove this code. If we merge this into `main` and release it, it should go to `amr-v1.3.2`.

rzlim08 (approved):
This looks good, I've tested this out on the failing samples and they seem to get the expected values.

PR #308, opened 2023-12-01
case insensitive phage match

ninabernick:
Previously we were only matching against "phage" but we want to count words like "Phage" as phage viruses as well. Uses `re.IGNORECASE` for case-insensitive full word match.

### DIFF
--- a/workflows/index-generation/generate_lineage_csvs.py
+++ b/workflows/index-generation/generate_lineage_csvs.py
@@ -38,6 +38,7 @@ _versioning_fieldnames = [
     "updated_at",
 ]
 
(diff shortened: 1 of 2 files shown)


rzlim08:
## Comments 
* Did a quick search of the Taxonomy browser and the match looks good

## How was this reviewed?
- [x] Read through the code
- [ ] Ran the code and examined output files
- [ ] Stepped through the code


## Are there files that were not reviewed?
- No

## Are there tests included
- [x] Yes
- [ ] No - and why?

rzlim08 (approved):

PR #313, opened 2024-01-16
[CZID-8743] Bulk download pipeline

robertaboukhalil:


morsecodist (approved):
Tested end to end and it works

PR #314, opened 2024-01-16
NCBI Compress - parallel split by taxid

phoenixAja:
* split NT and NR by taxid in parallel
* modify test data to only contain records with unique lengths
* updated index generation job to sort a directory of files rather than one file (now that we are breaking apart NT and NR in parallel into taxids sorting is not preserved)


morsecodist (approved):
Super good idea! Should be a nice speedup

PR #315, opened 2024-01-18
memory efficient loc db

morsecodist:
This makes generating the _loc and _info marisa tries much more memory efficient and maybe a little faster.

Both of these make an index on either nt or nr. _loc indexes from accession IDs to sequence offsets, _info indexes from accession IDs to accession names and sequence lengths. Previously these were combined so we could iterate through nr and nt once for both. However, this prevents us from using marisa trie's iterator constructor to save memory because we need two iterators with different data that will be read one after another.

By splitting these we do have to iterate twice but we only need to keep one row in memory at a time instead of both data sets entirely. Also we can do these in parallel because they use way less memory and only one core each.

As for why NR ran out of memory, this gets to the best bit. We actually don't use the info db for NR but we still built it because the script used both. This may be bigger with NR than NT because it stores the full names. If NR names are significantly longer for some reason it could lead to NR running out of memory before NT because the sequences are not loaded into memory at all. By completely dropping this we should have a major memory reduction for NR.


workflows/index-generation/generate_ncbi_db_index.py
```diff
@@ -0,0 +1,85 @@
...
+    # Sometimes multiple accessions will be mapped to a single sequence.
+    # In this case, they will be separated by the \x01 char.
+    # To get the accession name, just use the string until the first \x01.
+    parts = header_line.split('\x01', 1)[0].split(' ', 1)
+    accession_id = parts[0][1:]  # Remove the '>'
+    accession_name = parts[1] if len(parts) > 1 else ''
+    return accession_id, accession_name
+
+
+Accession = namedtuple('Accession', 'accession_id accession_name seq_offset header_len seq_len seq_bp_len')
```
phoenixAja, line 25: nit: I like the syntax where the field names are a list instead of a string
```suggestion
Accession = namedtuple('Accession', ['accession_id', 'accession_name', 'seq_offset', 'header_len', 'seq_len', 'seq_bp_len'])
```

workflows/index-generation/generate_ncbi_db_index.py
```diff
@@ -0,0 +1,85 @@
...
+            else:
+                seq_len += len(line)
+                seq_bp_len += len(line.strip())
+        if seq_len > 0 and len(accession_id) > 0:
+            yield Accession(accession_id, accession_name, seq_offset, header_len, seq_len, seq_bp_len)
+
+
+def generate_loc_db(db_file, loc_db_file):
+    loc_dict = marisa_trie.RecordTrie(
+        "QII",
```
phoenixAja, line 59: would be nice to have a comment on what this means

phoenixAja:
> As for why NR ran out of memory, this gets to the best bit. We actually don't use the info db for NR but we still built it because the script used both. This may be bigger with NR than NT because it stores the full names. If NR names are significantly longer for some reason it could lead to NR running out of memory before NT because the sequences are not loaded into memory at all. By completely dropping this we should have a major memory reduction for NR.

interesting! I did notice that pdb accession names are crazy long!

phoenixAja:
@morsecodist do we have test coverage for your changes? Otherwise i can make a follow up ticket to create some tests (and i can take that on)

workflows/index-generation/generate_ncbi_db_index.py
```diff
@@ -0,0 +1,85 @@
...
+        ((a.accession_id, (a.seq_offset, a.header_len, a.seq_len)) for a in _extract_accessions(db_file)),
+    )
+    loc_dict.save(loc_db_file)
+
+
+def generate_info_db(db_file, info_db_file):
+    info_dict = marisa_trie.RecordTrie(
+        "256pI",
```
phoenixAja, line 67: same here

workflows/index-generation/generate_ncbi_db_index.py
```diff
@@ -0,0 +1,85 @@
...
+    accession_id = parts[0][1:]  # Remove the '>'
+    accession_name = parts[1] if len(parts) > 1 else ''
+    return accession_id, accession_name
+
+
+Accession = namedtuple('Accession', ['accession_id', 'accession_name', 'seq_offset', 'header_len', 'seq_len', 'seq_bp_len'])
+
+
+def _extract_accessions(db_file: str):
+    with open(db_file) as dbf:
```
phoenixAja, line 29: nit: we could probably use the `SeqIO.FastaIO.FastaIterator(handle)` from biopython to iterate through fasta records in a more clear way. (so we could get the accession and sequence in a single record when we loop through)

phoenixAja (approved):
thank you!!

workflows/index-generation/generate_ncbi_db_index.py
morsecodist: I considered this but because we need to include newlines to get the exact byteranges as well as some other info it might cause problems. I also wanted to maintain the logic because of the low test coverage. I will add a comment explaining

phoenixAja: i have an example in my check_sequence_retention notebook [here](https://github.com/chanzuckerberg/czid-workflows/blob/tmorse-nt-compression/workflows/index-generation/ncbi-compress/check_sequence_retention.ipynb)

phoenixAja: ok sounds good!

PR #316, opened 2024-01-18
Add tracking for spades failures

lvreynoso:
[CZID-2148](https://czi-tech.atlassian.net/browse/CZID-2148?atlOrigin=eyJpIjoiNThjNTI0ZDQ5NjI4NDdiZWJiMzA3YzIzYTBjYzllM2IiLCJwIjoiaiJ9)

[CZID-2148]: https://czi-tech.atlassian.net/browse/CZID-2148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

This PR adds python code to collect data from SPAdes logs from assembly runs with no contigs and write that data to a json file, which is a WDL output.

The python script collects the contents of `warnings.log`. It also looks in `spades.log` for any errors that may appear, and a stack trace if it appears.

Since this script is used in both the short-read-mngs workflow and the amr workflow, I placed it in the lib/bin/ directory and included it in the short-read-mngs Dockerfile. This docker image is used by the RunSpades task in both workflows, which would seem to make the script available to both workflows; however, due to the fact that the amr workflow's tests for RunSpades rely on this docker image and in conjunction with our current setup, relying on new changes to this image in the same PR as the image is modified causes the tests to break. This is the reason the script code is inlined in the amr workflow.


scripts/release.sh
```diff
@@ -28,6 +28,14 @@ else
     exit 1
 fi
 
+if [[ $WORKFLOW_NAME == amr ]]; then
```
rzlim08, line 31: Can we just use regular tags for this?

lib/bin/log_assembly_fail.py
```diff
@@ -0,0 +1,35 @@
...
+SPADES_LOG = "assembly/spades/spades.log"
+
+failure_info = {
+    "log_present": False,
+    "warnings_present": False,
+    "warnings": None,
+    "stack_trace": None,
+    "errors": None,
+}
+
```
rzlim08, line 14: It would be good to wrap this in functions with an entrypoint

workflows/amr/run.wdl
```diff
@@ -616,31 +616,70 @@ task RunSpades {
...
-        else
-            spades.py -s ~{reduplicated_reads[0]} -o "spades/" -m 100 -t 36 --only-assembler 1>&2
-        fi
+        python3 <<CODE
+    import json
+    import os
+
+    WARNINGS_LOG = "spades/warnings.log"
+    SPADES_LOG = "spades/spades.log"
+
```
rzlim08, line 631: Could we reuse the script above?

rzlim08 (commented):
A broad question, do we know what we're parsing for here?

scripts/release.sh
lvreynoso: Oops, this file should not be committed, I'll fix that

workflows/amr/run.wdl
lvreynoso: I can try again. I ran into an issue where modifying the RunSpades task to rely on a change in the short-read-mngs container broke several AMR tests. I don't remember if this prevented a build and push of the new image, I will look into it again.

rzlim08 (approved):
Looks great! Thanks!

PR #317, opened 2024-01-18
similarity check rearrangement

phoenixAja:
* check for similarity in chunk before checking in tree
* current tests cover this change

### DIFF
--- a/workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
+++ b/workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
@@ -371,7 +371,7 @@ pub mod ncbi_compress {
             // create signatures for each record in the chunk
             let chunk_signatures = chunk
             .par_iter()
-            .filter_map(|r| {
+            .map(|r| {
                 let record = r.as_ref().unwrap();
                 let mut hash;
                 if is_protein_fasta {
@@ -384,7 +384,7 @@ pub mod ncbi_compress {
                         0,
                     );
                     hash.add_protein(record.seq()).unwrap();
-                    Some((hash, record))
+                    (hash, record)
                 } else {
                     hash = KmerMinHash::new(
                         scaled,
@@ -395,7 +395,7 @@ pub mod ncbi_compress {
                         0,
                     );
                     hash.add_sequence(record.seq(), true).unwrap();
-                    Some((hash, record))
+                    (hash, record)
                 }
             }).collect::<Vec<_>>();
 
--- a/workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
+++ b/workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
@@ -15,7 +15,6 @@ pub mod ncbi_compress {
     use sourmash::signature::SigsTrait;
     use sourmash::sketch::minhash::KmerMinHash;
 
-    use crate::fasta_tools::fasta_tools::sort_fasta_by_sequence_length;
     use crate::logging::logging;
     use crate::minhashtree_w_logging::minhashtree_w_logging::{
         MinHashTreeWithLogging, MinHashTreeWithLoggingFunctionality,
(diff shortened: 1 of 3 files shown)


workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
```diff
@@ -346,45 +345,87 @@ pub mod ncbi_compress {
...
+
+            // we need to make sure records within the chunk arn't similar to each other before
+            // we check them against the larger tree
+            for (hash, record) in chunk_signatures {
+                let similar = unique_in_chunk
+                    .par_iter()
+                    .any(|(other, _record)| containment(&hash, &other).unwrap() >= similarity_threshold);
+
+                if !similar {
+                    unique_accession_count.add_assign(1);
```
morsecodist, line 410: This may still end up being filtered out later we shouldn't increment. We should move this down to the write.

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
```diff
@@ -346,45 +345,87 @@ pub mod ncbi_compress {
...
-            .collect::<Vec<_>>();
-        accession_count.add_assign(chunk.len() as u64);
-        while chunk.len() > 0 {
-            // Take a chunk of accessions and return the ones that are not already in sketches
+
+
+        loop {
+            // Initialize a temporary vector to store the unique items from each chunk
+            let mut unique_in_chunk: Vec<(KmerMinHash, &fasta::Record)> = Vec::with_capacity(chunk_size);
+            let mut unique_in_tree_and_chunk: Vec<KmerMinHash> = Vec::with_capacity(chunk_size);
```
morsecodist, line 358: Slight performance thing by declaring these outside of the loop than clearing it will save you allocations.

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
```diff
@@ -346,45 +345,87 @@ pub mod ncbi_compress {
...
+                    hash.add_sequence(record.seq(), true).unwrap();
+                    Some((hash, record))
+                }
+            }).collect::<Vec<_>>();
+
+            // we need to make sure records within the chunk arn't similar to each other before
+            // we check them against the larger tree
+            for (hash, record) in chunk_signatures {
+                let similar = unique_in_chunk
+                    .par_iter()
```
morsecodist, line 406: I think I do this as well but it actually may not be worth the overhead to parallelize this with a chunk size of 1000.

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
```diff
@@ -399,44 +440,18 @@ pub mod ncbi_compress {
                     {
                         None
                     } else {
-                        Some((record, hash))
+                        let mut writer = writer.lock().unwrap();
```
morsecodist, line 443: I would avoid the mutex here and iterate through `unique_in_tree_and_chunk` sequentially and write them. The writes need to happen sequentially anyway so this probably doesn't buy you additional performance and could even be a hindrance.

This is also a bit awkward with `filter_map` which typically doesn't have side effects.

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
```diff
@@ -399,44 +440,18 @@ pub mod ncbi_compress {
...
             //     tree.insert(hash);
             // }
-            sketches.extend_from_slice(&tmp);
-            tmp.clear();
-            chunk = records_iter
-                .borrow_mut()
-                .take(chunk_size)
-                .collect::<Vec<_>>();
+
+            sketches.extend_from_slice(unique_in_tree_and_chunk.as_slice());
```
morsecodist, line 454: You could add the iteration and writing just above this line.

morsecodist (changes requested):

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
```diff
@@ -346,45 +345,87 @@ pub mod ncbi_compress {
...
-                    } else {
-                        hash = KmerMinHash::new(
-                            scaled,
-                            k,
-                            HashFunctions::murmur64_DNA,
-                            seed,
-                            false,
-                            0,
+            .par_iter()
+            .filter_map(|r| {
```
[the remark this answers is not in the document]
phoenixAja: thanks for catching that

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
phoenixAja: true, i moved it to a new block!

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
phoenixAja: i don't think it hurts too much, also thinking that maybe we can bump up the chunk size to be more

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
phoenixAja: fixed

workflows/index-generation/ncbi-compress/src/ncbi_compress.rs
phoenixAja: fixed

morsecodist (approved):
looks good!

PR #320, opened 2024-01-26
containment check perf boost

morsecodist:
Comment says it all, this should speed things up without a cost really.


phoenixAja (approved):
looks good! thanks!

PR #322, opened 2024-02-06
Pin goimports and gopy versions

lvreynoso:
Fixes an issue where new builds of the short-read-mngs Docker image fail if they aren't using cached versions of the s3quilt stage due to updated dependencies in the latest versions of goimports and gopy.


lvreynoso:
After discussion between @morsecodist we have decided to merge this PR despite the failing short-read-mngs viral benchmarks / run-samples CI action since we could not reproduce the failure locally. Pulling http://ghcr.io/chanzuckerberg/czid-workflows/czid-short-read-mngs-public:56fbf4b and running the test locally worked.

PR #345, opened 2024-04-11
add accession info to manifest

rzlim08:
* Adds accession info and taxon info into cg manifest. Doesn't do anything with it.

### DIFF
--- a/workflows/consensus-genome/manifest.yml
+++ b/workflows/consensus-genome/manifest.yml
@@ -45,6 +45,26 @@ raw_inputs:
     description: CZID's snapshot version of NCBI's index i.e. '2021-01-22'
     type: str
     required: True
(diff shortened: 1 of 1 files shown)


workflows/consensus-genome/manifest.yml
```diff
@@ -45,6 +45,26 @@ raw_inputs:
...
+    name: Accession Name
+    description: accession name
+    type: str
+    required: False
+  accession_id:
+    name: Accession ID
+    description: accession id
+    type: str
+    required: False
+  taxon:
```
j-x-han, line 58: I'm using `taxon_name` to be a bit more specific, could go either way though!

j-x-han (approved):
Overall LGTM, just one note

PR #360, opened 2024-05-20
check for completed alignment chunks before dispatching them

morsecodist:
Currently, if an alignment coordinator job fails due to a spot interruption the next job reruns all of the alignment chunks. It is likely that chunks will be complete, especially since the jobs are not cancelled and will continue working between retries. This can prevent a lot of redundant work and keep our alignment job queue smaller which should slightly mitigate other issues.

### DIFF
--- a/lib/idseq_utils/idseq_utils/batch_run_helpers.py
+++ b/lib/idseq_utils/idseq_utils/batch_run_helpers.py
@@ -33,7 +33,9 @@ config = Config(
 
 
 _batch_client = boto3.client("batch", config=config)
-_s3_client = boto3.client("s3")
+# the retries are less necessary for S3 because rate limiting is more generous but we have hit the limit
+#   and it is costly for this job to be re-run
+_s3_client = boto3.client("s3", config=config)
 
 try:
     account_id = boto3.client("sts").get_caller_identity()["Account"]
@@ -203,6 +205,7 @@ def _run_chunk(
         log.info(f"skipping chunk, output already exists: {wdl_output_uri}")
         return
     except ClientError as e:
+        # raise the error if it is anything other than "not found"
         if e.response["Error"]["Code"] != "404":
             raise e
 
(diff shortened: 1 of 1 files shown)


lib/idseq_utils/idseq_utils/batch_run_helpers.py
```diff
@@ -189,10 +189,23 @@ def _job_queue(provisioning_model: str):
...
     wdl_workflow_uri = f"s3://idseq-workflows/{aligner}-{aligner_wdl_version}/{aligner}.wdl"
 
-    input_bucket, input_key = _bucket_and_key(wdl_input_uri)
+    # if this job fails we don't want to re-run chunks that have already been processed
+    #   the presence of the output file means the chunk has already been processed
+    try:
+        _s3_client.head_object(Bucket=output_bucket, Key=output_key)
+        log.info(f"skipping chunk, output already exists: {wdl_output_uri}")
+        return
+    except ClientError as e:
```
rzlim08, line 205: I think we would want a `pass` here since we don't want to raise errors if the object doesn't exist, we just want to run the chunk. Is `ClientError` exhaustive as far as the errors that could come out of `.head_object`?

morsecodist: This behavior passes if the error is a `404`. The current behavior raises all non-404 client errors. So if there were an unrelated error encountered while doing the check it would be raised. I can see maybe we would want to ignore those because we could recover by re-running the chunk but I think they will be notable enough that we would want to know and fix the issue.

rzlim08: Sorry I saw `==` instead of `!=` there. Would errors from s3 throttling be raised here?

morsecodist: Yeah it's confusing. Adding a comment. Throttle issues would be thrown here but we are already doing a put or two per chunk so the behavior is similar. We seem to mostly get those very rarely during the polling phase. I can add the adaptive strategy with lots of retries that we use for batch to the S3 client.

rzlim08 (approved):
Thanks!

lib/idseq_utils/idseq_utils/batch_run_helpers.py
rzlim08: Great thanks!

PR #365, opened 2024-05-23
Sort bowtie2 outputs for reproducibility

rzlim08:
* This PR attempts to make almost all outputs of the short-read-mngs pipeline reproducible when run on the same input. 
* The changes made here are issues in sorting: 
1. sort bowtie2_ercc.sam file 
2. sort bowtie2 output from reads to contig mapping
3. sort list of NT/NR source in output json

The outputs were tested as following:
1. Run 2 pipelines with the same input
2. Run `python3 scripts/compare-outputs.py <output.json> <output2.json>` to compare md5 hashes
3. Manually `diff` files to determine the cause of the discrepancies

The remaining files which are different on every run are: 
czid_short_read_mngs.host_filter.fastp_html
czid_short_read_mngs.postprocess.assembly_out_assembly_spades_output_log

Both of which contain timestamps which account for the discrepancy

### DIFF
--- a/lib/idseq-dag/idseq_dag/steps/run_assembly.py
+++ b/lib/idseq-dag/idseq_dag/steps/run_assembly.py
@@ -237,7 +237,7 @@ class PipelineStepRunAssembly(PipelineStep):
                 args=samtools_sort_params
             )
         )
-        
+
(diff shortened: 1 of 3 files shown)


lvreynoso (approved):
looks good! great work!

PR #379, opened 2024-10-01
[Bugfix] Fix subsample depth

rzlim08:
* reduce subsampling depth


vincent-czi (approved):
Code tweak seems reasonable. Might be nice to just comment in this PR a little more info of what was going on before, maybe reference the ticket if there was one? But code is certainly approved!

